PR #7, opened 2018-09-17
Example for incremental metrics computation

sscdotopen:
Added an example for the stateful incremental metrics computation


src/main/scala/com/amazon/deequ/examples/Item.scala
```diff
@@ -0,0 +1,9 @@
+package com.amazon.deequ.examples
```
andrey-tpt: Copyright ?

sscdotopen: good point, will update!

PR #12, opened 2018-09-17
Remove Imputer

tdhd:
*Description of changes:*

Did we want to keep the imputer in this package?


sscdotopen:
I'm a little torn about the imputer. On the one hand, imputation is not really the goal of this library and with the imputer, we could also remove the dependency on SparkML. On the other hand, without the imputer, we cannot provide the "predicability" metric that we have been planning.

I would be ok with removing it for now, to keep the package lean, we can always reactivate it later on. What do others think here?

tdhd:
Just checked the travis logs, we can't remove `mllib` because of some breeze imports by the anomaly detection code: `import breeze.stats.meanAndVariance`. I think we should still drop the mllib dependency and adapt the code a bit.

Same thoughts here regarding predictability, if we want it back we can revive it.

sscdotopen:
we can also add a dependency on breeze alone, that should be fine.

tdhd:
Good point, added breeze now

dustin-lange:
Good idea to remove the imputer as long as it is not needed.

Can we run a check if other files live in isolation as well?

sscdotopen:
We already removed quite a lot of files. Another candidate for removal would be the row-level schema validation which is kind of isolated from the rest of the codebase.

dustin-lange:
Removing row-level validation could make sense. Or we should think about integrating it with the main API?

sscdotopen:
Its hard to integrate into the rest of the code because it works on row-level while all the other parts look at the table level only.

PR #31, opened 2018-09-25
Fixed broken Constraint Suggestion Link in Readme

stefan-grafberger:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Our library contains much more than what we showed in the basic example. We will
  
  * [Anomaly detection](src/main/scala/com/amazon/deequ/anomalydetection)
  * Single-column [data profiling](https://github.com/awslabs/deequ/tree/master/src/main/scala/com/amazon/deequ/profiles)
- * Automatic [suggestion of constraints](src/main/scala/com/amazon/deequ/suggestions/EndToEndConstraintSuggestion.scala)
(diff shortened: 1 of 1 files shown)


src/test/scala/com/amazon/deequ/suggestions/ConstraintSuggestionsIntegrationTest.scala
```diff
@@ -47,10 +47,10 @@ class ConstraintSuggestionsIntegrationTest extends WordSpec with SparkContextSpe
       val categories = Array("DE", "NA", "IN", "EU")
 
       val records = (0 until numRecords)
-        .map { _ =>
+        .map { record =>
```
sscdotopen: this is not required as the chance for a collision in UUID is really low

stefan-grafberger: I found that on some occasions (needed 6 tries just now to reproduce this) the line where we assert that we suggest an isUnique constraint fails. The explanation I could imagine for that is that the randomness combined with the way we calculate the approximate Uniqueness using the HLL sketch could be a problem. So there aren't actual collisions but the approximate Uniqueness is not accurate enough and too low. Is that possible?

PR #32, opened 2018-09-26
 Add markdown documention for the metrics repository example #27 

sscdotopen:
*Issue #27*


stefan-grafberger:
Maybe also include how to use .reuseExistingResultsForKey(resultKey, failIfResultsForReusingMissing=true/false)?

stefan-grafberger:
And note that these commands are supported by all 4 of our main entry points: AnalysisRunner, VerificationSuite, ColumnProfilerRunner and ConstraintSuggestionRunner

sscdotopen:
These are good points, but I think they are already very advanced and should better be covered in a different example.

Issue #9, opened 2018-09-17
Create an example for data profiling


sscdotopen:
Create an example for using our [single-column data profiling](src/main/scala/com/amazon/deequ/suggestions/ColumnProfiler.scala) to profile data.

hangelwen:
+1 for this. I'm exploring this lib for our data quality monitoring use case. Could you add some examples for this? Thanks.

sscdotopen:
Hi, we're happy to hear that your exploring the use of deequ! We'll provide you with this example soon.

PR #34, opened 2018-09-27
Add markdown documention for the data profiling example

sscdotopen:
*Issue #9*


sscdotopen:
It would be great to have a second example that shows the advanced options and gives hints on visualization. Could you create an issue for that? 

Where exactly should we link add links? Each example always links to its scala file and we have a link to the example package in README.md already.

stefan-grafberger:
Maybe something like this? However, even without this, it should be quite easy to find these classes in the code. 

We have two different types of profiles we return, [StandardColumnProfiles](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/profiles/ColumnProfile.scala) and [NumericColumnProfiles](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/profiles/ColumnProfile.scala). If the cardinality of a column is low enough, we also include a histogram with the StandardColumnProfile.

We also have more advanced configuration options as part of the ColumnProfilerRunner API, see [here](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/profiles/ColumnProfilerRunBuilder.scala). We will also include an example of a more advanced use case in the future.

Issue #54, opened 2018-09-30
Support Spark 2.3


hangelwen:
Hi Deequ team,

This is a pretty useful tool. However, it only supports Spark 2.2. Is there a plan to support Spark 2.3 version?

hangelwen:
@sscdotopen thanks for the quick reply. 

I tried and it does not work on Spark 2.3.  One issue I found it that some variables defined for HyperloglogPlus was moved from HyperLogLogPlusPlus to HyperLogLogPlusPlusHelper in this [PR ](https://github.com/apache/spark/commit/1d1a09be9fe07244d7d12442f6105f823b260755)

https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/catalyst/StatefulHyperloglogPlus.scala#L195
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/HyperLogLogPlusPlusHelper.scala#L280

sscdotopen:
Thank you for the hint, I did not think about the fact that we also extended some internal functions. I'll file this issue as a bug on our side, and we will have a look into that.

sscdotopen:
@hangelwen We think we have fixed the incompatibility, we will release a new RC later today.

PR #57, opened 2018-10-01
Support Spark 2.3

sscdotopen:
*Issue #54 *

Spark 2.3 contains some incompatible internal changes in catalyst, which made it hard for us to support both Spark 2.2 and Spark 2.3. I found some work arounds which required a little bit of code copying and some invocations via reflection. 

I could build a jar with 2.2 as dependency which worked on 2.2 and 2.3 clusters in EMR (I tested column profiling which uses the problematic aggregation functions), so I am quite confident that we should support Spark 2.3 now.


groodt:
Should the docs be updated accordingly that it now supports Spark 2.3? I don't suppose anybody has tried on Spark 2.4?

Issue #10, opened 2018-09-17
Create an example for constraint suggestion


sscdotopen:
Create an example for computing and evaluating [constraint suggestions](src/main/scala/com/amazon/deequ/suggestions/EndToEndConstraintSuggestion.scala) for data.

PR #61, opened 2018-10-04
Example for constraint suggestion

sscdotopen:
*Issue #10*


sscdotopen:
I also think that we should have an example with evaluation of the constraints. I wanted to initially include it, but it gives very weird results for a mini toy dataset, as the evaluation will be run on only 2 or 3 records. Could you create an issue for the extended example?

tdhd:
Added https://github.com/awslabs/deequ/issues/62

PR #66, opened 2018-10-11
non negative/is positive checks + constraint suggestion changes

tdhd:
*Description of changes:*

Looked into a use case of constraint suggestion today and found that the `isNonNegative` and `isPositive` constraints were generating misleading compliance metric values as they included `null` rows in the computation (e.g. 10% non negative compliance in a column with 10% positive non-nulls). Instead, a Compliance of 100% should be reported.

**Changes**:

* disregard `null`s in non negative and is positive constraints
* also suggest non negative constraint for a column if the minimum is `>= 0` and not only when the minimum of a column is 0
* previous change led to overlap of `isPositive` and `nonNegative` rules, removed positive constraint suggestion rule
* adapt and add tests


src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -672,7 +672,7 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
-    satisfies(s"$column >= 0", s"$column is non-negative", hint = hint)
+    satisfies(s"COALESCE($column, 0.0) >= 0", s"$column is non-negative", hint = hint)
```
dustin-lange: Can you add comments that describe why you are using COALESCE? That will be helpful for other readers of the code who haven't experienced the problems with missing values.

src/main/scala/com/amazon/deequ/suggestions/rules/NonNegativeNumbersRule.scala
```diff
@@ -26,8 +26,7 @@ case class NonNegativeNumbersRule() extends ConstraintRule[ColumnProfile] {
 
   override def shouldBeApplied(profile: ColumnProfile, numRecords: Long): Boolean = {
     profile match {
-      case numericProfile: NumericColumnProfile
-        if numericProfile.minimum.isDefined && numericProfile.minimum.get == 0.0 => true
+      case numericProfile: NumericColumnProfile => numericProfile.minimum.exists(_ >= 0.0)
```
dustin-lange: Is this merely a simplification of the code or a change in functionality?

tdhd: This changes the behavior such that non negative constraints should be applied if the minimum of a column is >= 0.

src/main/scala/com/amazon/deequ/checks/Check.scala
tdhd: Yes, good point, will add.

Issue #16, opened 2018-09-18
Verify license headers automatically


sscdotopen:
We should configure the [RAT plugin](http://creadur.apache.org/rat/apache-rat-plugin/) for maven and have it check that all .scala files have appropriate license headers.

PR #70, opened 2018-10-20
Verify license headers with Rat

strongh:
Addresses #16 

Added Apache Rat Maven plugin to verify the presence of license headers in Scala files, and added headers where they were missing.

+ Verification runs automatically at compile time.
+ Only Scala files are checked.
+ The default Apache 2.0 matcher didn't work because it expects a specific [URL](https://github.com/apache/creadur-rat/blob/ecbc6605f8e613b7ce52195439bcccde97e28ae9/apache-rat-core/src/main/java/org/apache/rat/analysis/license/ApacheSoftwareLicense20.java#L29). I added [simple customer](http://creadur.apache.org/rat/apache-rat-plugin/examples/custom-license.html) license matcher instead on [these](https://github.com/strongh/deequ/blob/rat-verify/pom.xml#L219-L222) patterns.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,7 +111,6 @@
     </dependencies>
 
 
-
     <build>
         <sourceDirectory>src/main/scala</sourceDirectory>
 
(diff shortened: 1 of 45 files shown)


sscdotopen:
Thank you very much Homer!

PR #74, opened 2018-10-29
applicability for byte, short and long data types

tdhd:

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
@@ -31,6 +31,90 @@ private[deequ] case class ApplicabilityResult(
   failures: Seq[(String, Throwable)]
 )
 
+private[deequ] object Applicability {
+  def randomString(nullable: Boolean = false): java.lang.String = {
+    if (nullable && math.random < 0.01) {
+      null
+    } else {
+      val length = (math.random * 20 + 1).toInt
+      RandomStringUtils.randomAlphanumeric(length)
+    }
+  }
+
+  def randomInteger(nullable: Boolean = false): java.lang.Integer = {
+    if (nullable && math.random < 0.01) {
+      null
+    } else {
+      (math.random * 100).toInt
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
```diff
@@ -31,13 +31,98 @@ private[deequ] case class ApplicabilityResult(
...
+  def randomString(nullable: Boolean = false): java.lang.String = {
+    if (nullable && math.random < 0.01) {
+      null
+    } else {
+      val length = (math.random * 20 + 1).toInt
+      RandomStringUtils.randomAlphanumeric(length)
+    }
+  }
+
+  def randomInteger(nullable: Boolean = false): java.lang.Integer = {
```
sscdotopen: we should refactor the random number generation out of all the typed methods into a single place and only invoke the conversion in the actual methods.

src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
```diff
@@ -31,13 +30,82 @@ private[deequ] case class ApplicabilityResult(
   failures: Seq[(String, Throwable)]
 )
 
+private[deequ] object Applicability {
+
+  private def shouldBeNull(nullable: Boolean): Boolean = {
+    nullable && math.random < 0.01
+  }
+
+  private def nullableIf(nullable: Boolean)(n: java.lang.Number): java.lang.Number = {
```
sscdotopen: I think this is a bit problematic, as all methods now return java.lang.Number instead of their designated type. Why not simply return a double here and then cast it in the invoking methods?

tdhd: Using `Number` still allows spark to create the correct schema for the random data. Am I missing something else?

sscdotopen: I did not mean that its incorrect, I just think it makes the code harder to read, to understand and to debug. If the return type is Number, than a method which should return a Byte could return a Double with a compiler error. We should not rely on the ability of the invoking code to correctly cast the result.

PR #75, opened 2018-10-31
Exceptions instead of Throwables

tdhd:
*Description of changes:*

Use `Exception` instead of `Throwable`.


sscdotopen:
Whats the reasoning here? I think we explictly made this catch all throwables so that no error is  propagated to the invoking code.

tdhd:
Apart from a few corner cases `Throwable`s shouldn't be caught. When catching `Throwable`s we include `Exception` and `Error` where `Error` usually is not recoverable from, like `VirtualMachineError` or `OutOfMemoryError`, i.e. we want the invoking code to fail.

src/main/scala/com/amazon/deequ/analyzers/runners/MetricCalculationException.scala
```diff
@@ -21,12 +21,12 @@ abstract class MetricCalculationException(message: String) extends Exception(mes
...
 
-  def this(message: String, cause: Throwable) {
+  def this(message: String, cause: Exception) {
     this(message)
     initCause(cause)
   }
 
-  def this(cause: Throwable) {
+  def this(cause: Exception) {
     this(Option(cause).map(_.toString).orNull, cause)
```
sscdotopen: We should have better logic for the case where cause can be null. We should not pass the nulls to the constructor.

tdhd: When are we expecting the cause to be null?

tdhd: Yes, I was wondering why we are explicitly handling the case when `cause` can be null. Are we expecting it to be null at any point?

sscdotopen: No, not really. We could check it and invoke the constructor with a custom runtime exception to be super-duper safe, but we should not propagate nulls.

PR #78, opened 2018-12-04
Fix random value generation for decimals during applicability checks

sscdotopen:
In the current, applicability checks, Deequ has a bug in generating random number for decimals. If the decimal has low precision we accidentally generate a column of nulls (as the random numbers fall outside of the range of the decimal), which causes the applicability check to crash.

This PR fixes the issue by generating decimals appropriately corresponding to the specified scale and precision. 

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
@@ -25,7 +25,7 @@ import com.amazon.deequ.metrics.Metric
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.{DataFrame, Row, SparkSession}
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/amazon/deequ/analyzers/applicability/Applicability.scala
```diff
@@ -195,6 +218,8 @@ private[deequ] class Applicability(session: SparkSession) {
 
     val data = generateRandomData(schema, 1000)
 
+    data.select("decimalCol3").head(10).foreach { println }
```
tdhd: Remove please

sscdotopen: Oops...

PR #83, opened 2019-01-28
Add timerange containment check

iamsteps:
Adds isContainedIn check for timestamp ranges.

### DIFF
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -848,7 +848,7 @@ case class Check(
     * @param column column to run the assertion
     * @param lowerBound lower bound of the interval
     * @param upperBound upper bound of the interval
-    * @param includeLowerBound is a value equal to the lower bound allows?
+    * @param includeLowerBound is a value equal to the lower bound allowed?
     * @param includeUpperBound is a value equal to the upper bound allowed?
     * @param hint A hint to provide additional context why a constraint could have failed
     * @return
@@ -870,7 +870,7 @@ case class Check(
     * @param column column to run the assertion
     * @param lowerBound lower bound of the interval
     * @param upperBound upper bound of the interval
-    * @param includeLowerBound is a value equal to the lower bound allows?
+    * @param includeLowerBound is a value equal to the lower bound allowed?
     * @param includeUpperBound is a value equal to the upper bound allowed?
     * @param hint A hint to provide additional context why a constraint could have failed
     * @param quoteBounds should the bound values be quoted?
@@ -905,7 +905,7 @@ case class Check(
     * @param column column to run the assertion
     * @param lowerBound lower bound of the interval
     * @param upperBound upper bound of the interval
-    * @param includeLowerBound is a value equal to the lower bound allows?
+    * @param includeLowerBound is a value equal to the lower bound allowed?
     * @param includeUpperBound is a value equal to the upper bound allowed?
     * @param hint A hint to provide additional context why a constraint could have failed
     */
@@ -945,7 +945,7 @@ case class Check(
     * @param column column to run the assertion
     * @param lowerBound lower bound of the interval
     * @param upperBound upper bound of the interval
-    * @param includeLowerBound is a value equal to the lower bound allows?
+    * @param includeLowerBound is a value equal to the lower bound allowed?
     * @param includeUpperBound is a value equal to the upper bound allowed?
     */
   def isContainedIn(
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -873,7 +873,8 @@ case class Check(
     * @param includeLowerBound is a value equal to the lower bound allowed?
     * @param includeUpperBound is a value equal to the upper bound allowed?
     * @param hint A hint to provide additional context why a constraint could have failed
-    * @param quoteBounds should the bound values be quoted?
+    * @param quoteBounds should the bound values be quoted? Quoting is required for data
+             types such as DateType, TimestampType, StringType.
     */
   private def isContainedInRange[T](
       column: String,
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -25,6 +25,7 @@ import com.amazon.deequ.metrics.{Distribution, Metric}
 import com.amazon.deequ.repository.MetricsRepository
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import com.amazon.deequ.anomalydetection.HistoryUtils
+import java.sql.Timestamp
 import scala.util.matching.Regex
 
 object CheckLevel extends Enumeration {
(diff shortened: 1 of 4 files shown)


src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -860,16 +861,120 @@ case class Check(
...
+  /**
+    * Asserts that the non-null values in a column fall into the predefined interval
+    *
+    * @param column column to run the assertion
+    * @param lowerBound lower bound of the interval
+    * @param upperBound upper bound of the interval
+    * @param includeLowerBound is a value equal to the lower bound allows?
+    * @param includeUpperBound is a value equal to the upper bound allowed?
+    * @param hint A hint to provide additional context why a constraint could have failed
+    * @param quoteBounds should the bound values be quoted?
```
dustin-lange: when/why should they be quoted?

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -860,16 +861,120 @@ case class Check(
...
+      isContainedInRange(column, lowerBound, upperBound, includeLowerBound, includeUpperBound, hint)
+  }
 
+  /**
+    * Asserts that the non-null values in a column fall into the predefined interval
+    *
+    * @param column column to run the assertion
+    * @param lowerBound lower bound of the interval
+    * @param upperBound upper bound of the interval
+    * @param includeLowerBound is a value equal to the lower bound allows?
```
dustin-lange: allow[ed]

or: whether the lower bound is included in the interval

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -860,16 +861,120 @@ case class Check(
...
+
+  /**
+    * Asserts that the non-null values in a timestamp column fall into the predefined interval
+    *
+    * @param column column to run the assertion
+    * @param lowerBound lower bound of the interval
+    * @param upperBound upper bound of the interval
+    * @param hint A hint to provide additional context why a constraint could have failed
+    */
+  def isContainedIn(
```
dustin-lange: Why are there 3 variations of the function required? Is it for the override?

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -860,16 +861,120 @@ case class Check(
...
+    * @param column column to run the assertion
+    * @param lowerBound lower bound of the interval
+    * @param upperBound upper bound of the interval
+    * @param includeLowerBound is a value equal to the lower bound allows?
+    * @param includeUpperBound is a value equal to the upper bound allowed?
+    * @param hint A hint to provide additional context why a constraint could have failed
+    */
+  def isContainedIn(
+      column: String,
+      lowerBound: Timestamp,
```
dustin-lange: I can imagine that some people wouldn't want to define a java.sql.Timestamp. Can they also filter using a string?

For example, in Postgres/Redshift SQL, you can define something like "date >= '2013-01-01" or "date between '2013-01-01' and '2014-01-01'".

How could we support a use case like "the date should not be in the future" or "the date should be within the last 3 days"?

src/main/scala/com/amazon/deequ/checks/Check.scala
iamsteps: The data types that need quoting are a superset of `{ DateType, TimestampType, StringType }`. These types need quoting when used in a query as below.

dustin-lange: Can you add this information to the comment?

src/main/scala/com/amazon/deequ/checks/Check.scala
iamsteps: Makes sense. Some suggestions:

- we could add a dependency on [scala-time](https://github.com/jorgeortiz85/scala-time/tree/master) / [Joda-time](https://github.com/JodaOrg/joda-time), and accept `DateTime` objects in the `isContainedIn` function. This would allow richer interval specifications directly, such as 

  ```
  isContainedIn("column", DateTime.now(), DateTime.now() - 3.days())  
  ```
  or

  ```
  val now = DateTime.now()
  isContainedIn("column", now, now.minus(Period.days(3)))
  ```

  This would also allow users to specify

  ```
  isContainedIn("column", DateTime.parse("2013-01-01"), DateTime.parse("2014-01-01"))
  ``` 
  which is pretty close to directly supplying `String`s as timestamps.

- in the first version of this PR, I used `java.sql.Timestamp` because of the [direct mapping](https://jaceklaskowski.gitbooks.io/mastering-spark-sql/spark-sql-DataType.html#TimestampType) to Spark's `TimestampType`.

- We could support specification of `String` ranges, in which case the users could directly input bounds such as `"2013-01-01"`. However, such a function could be used for arbitrary `String` ranges, such as `isContainedIn("column", "a", "z")`, which might have unintuitive behavior. If we were to support `String`s as timestamps directly, we might want to provide a dedicated name such as `isContainedInTimeRange`. 


I am favoring the approach to not support date/time Strings but instead provide an interface to `DateTime`, because valid formatting options are explicit by the use of Joda and permissible formats can be directly looked up and are standardized.

iamsteps: It makes sense but I have one remark. To support timestamp strings, I recommend to bring in the Joda dependency and let the string-typed method act as a wrapper around the `DateTime`-typed method, like so:

```scala
def isContainedInTimeRange(
    column: String,
    lowerBound: String,
    upperBound: String,
    ...) =
  isContainedInTimeRange(
    column,
    DateTime.parse(lowerBound),
    DateTime.parse(upperBound), 
    ...
  )
```

dustin-lange: Is that because there is no equivalent nice parse method from the Java SQL API?

One more remark on the parameter names: Should we call them `from` and `to` rather than `lowerBound` and `upperBound`?

iamsteps: - Regarding Joda dependency: yes, you are more restricted when using `java.sql.Timestamp`.  Timestamp format must be `yyyy-mm-dd hh:mm:ss[.fffffffff]`, so you cannot simply supply `"2019-01-01"` (which works for Joda's `DateTime`).

- Regarding parameter names: agree, since we will rename to `isContainedInTimeRange`, renaming `lowerBound` to `from` and `upperBound` to `to` seems like a good choice.

dustin-lange: Okay, let's postpone the Joda import and String variant for now. Maybe we can use it through the JDK8-included Joda/datetime version though?

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -860,16 +861,121 @@ case class Check(
...
+  private def isContainedInRange[T](
+      column: String,
+      lowerBound: T,
+      upperBound: T,
+      includeLowerBound: Boolean,
+      includeUpperBound: Boolean,
+      hint: Option[String] = None,
+      quoteBounds: Boolean = false)
+  : CheckWithLastConstraintFilterable = {
     val leftOperand = if (includeLowerBound) ">=" else ">"
```
sscdotopen: Please use Spark's functions to programmatically build the query rather than building it yourself using string concatenation.

src/main/scala/com/amazon/deequ/checks/Check.scala
sscdotopen: I would prefer the string variant as it is the easiest option for users. We can manually check for strings in the yyyy-mm-dd  format and expand them to the format that Timestamp expects.

sscdotopen:
I am closing this due to inactivity. Feel free to reopen later.

Issue #84, opened 2019-01-31
isContainedIn Check breaks for columns with special characters


malcolmgreaves:
The `isContainedIn` `Check`s will fail whenever there exists at least one column with a special character. Here, a special character means something that is reserved in the Spark SQL language: e.g. a `[` or `]`. `isContainedIn` generates SQL, but the column name is not properly escaped. This means that, at execution time, the generated SQL will fail with a syntax error. As a consequence, this syntax error will cascade through all other `Check`s when associated with a `VerificationRunBuilder`. Thus, even valid `Check`s will fail due to this SQL syntax error.

PR #85, opened 2019-01-31
Bugfix: Escape Column Name in Generated Spark SQL

malcolmgreaves:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/84

*Description of changes:*
This PR properly escapes column names in generated Spark SQL code.

Currently, when executing a `isContainedIn` `Check` on any `DataFrame` that contains any column whose name contains any characters that make up any reserved keyword in Spark SQL will cause a runtime failure. Digging into it, the runtime failure will show the root cause is a Spark SQL syntax error (specifically a `org.apache.spark.sql.catalyst.parser.ParseException`) due to incorrectly parsing the column name. 

E.g. if the column name starts with `[`, then the exception will read as:
```
org.apache.spark.sql.catalyst.parser.ParseException:
extraneous input '[' expecting ...
```

The fix for this bug is to wrap all bare column names in generated Spark SQL with backticks: \`\`. I.e. a query such as `$column IS NULL` is unsafe, but it's escaped variant is ok to execute, regardless of the value of `$column`:
```
`$column` IS NULL
```

A new test showcasing this functionality is in the `CheckTest` `class` named `"Check on column names with special characters"`. This test executes the `isContainedIn``Check` on both a value list as well as a (minimum, maximum) boundary. It succeeds only when the associated generated Spark SQL code has the escaped column name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.iml
 **/*.iml
 target/.travis/public-signing-key.gpg
+target/
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -837,7 +837,7 @@ case class Check(
       .map { _.replaceAll("'", "''") }
       .mkString("'", "','", "'")
 
-    val predicate = s"$column IS NULL OR $column IN ($valueList)"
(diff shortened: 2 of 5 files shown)


malcolmgreaves:
:D

Thank you for accepting & merging! `deequ` is wonderful!

Issue #89, opened 2019-02-08
>,<,>=,<= checks fail on column names with special characters


malcolmgreaves:
The `.isGreaterThan`, `.isLessThan`, `.isGreaterThanOrEqualTo`, and `.isLessThanOrEqualTo` methods on the `Check` type will fail with a Spark SQL `SyntaxError` at runtime when applied to columns whose names contain special characters or keywords.

malcolmgreaves:
```scala
// ensure that `spark` exists as an instance of `SparkSession`

import com.amazon.deequ.analyzers.Analysis
import com.amazon.deequ.analyzers.runners.AnalysisRunner
import com.amazon.deequ.checks.{ Check, CheckLevel, CheckResult } 

case class Wpair(`[this] is a value!!!11!#$`: Long, `[ also a column name:`: Long)
val rawValues = Seq(Long.MaxValue, Long.MaxValue, 25252L, 25252L, 66231L)
val weirdNums = spark.createDataFrame(
    rawValues.zip(rawValues.map{_-10L}).map { (Wpair.apply _).tupled }
)

val escapedCheck = Check(CheckLevel.Error, """escaped isGreaterThanOrEqualTo""").isGreaterThanOrEqualTo("`[this] is a value!!!11!#$`", "`[ also a column name:`")
val nonEscapedCheck = Check(CheckLevel.Error, """non-escaped isGreaterThanOrEqualTo""").isGreaterThanOrEqualTo("[this] is a value!!!11!#$", "[ also a column name:")

def analyze(df: DataFrame, check: Check): CheckResult = 
  check.evaluate(
    AnalysisRunner
      .onData(df)
      .addAnalyzers(Analysis().analyzers ++ check.requiredAnalyzers())
      .run()
    )

println(s"Check on weird column name, escaped:     ${analyze(weirdNums, escapedCheck).status}")
println(s"Check on weird column name, non-escaped: ${analyze(weirdNums, nonEscapedCheck).status}")
```

Observed output is:
```
Check on weird column name, escaped:     Success
Check on weird column name, non-escaped: Error
```

Investigating the `nonEscapedCheck`'s application, we have:
```scala
print(analyze(weirdNums, nonEscapedCheck).constraintResults.head.message.get)
```
Showing:
```
org.apache.spark.sql.catalyst.parser.ParseException: 
[...]

PR #91, opened 2019-02-14
Sanitize column names in spark sql calls

malcolmgreaves:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/89

*Description of changes:*
#### Problem
Deequ does not sanitize column names when creating Spark SQL statements.
Thus, deequ fails on `DataFrame`s that contain any special characters 
that are reserved in the Spark SQL syntax.

#### Tests Showing Failure
Tests have been changed to use at least one column name with special
characters. Specifically, all of the `att1` columns have been changed to
the value `]att1[`. When used in Spark SQL statements, these special
characters will, if not properly escaped, cause syntax errors during
interpretation.

This column name change has caused `com.amazon.deequ.checks.CheckTest`
to fail. Specifically, the failing tests are:
```
Check
...
- should correctly evaluate convenience constraints *** FAILED ***
  Error did not equal Success (CheckTest.scala:835)

Check on column names with special characters
...
- should work for isPositive *** FAILED ***
  Error did not equal Success (CheckTest.scala:885)
- should work for isNonNegative *** FAILED ***
  Error did not equal Success (CheckTest.scala:885)

Checks for two-columned DataFrames
- should check greater than *** FAILED ***
  Error did not equal Success (CheckTest.scala:885)
- should check less than *** FAILED ***
  Error did not equal Success (CheckTest.scala:885)
- should check greater than or equal to *** FAILED ***
  Error did not equal Success (CheckTest.scala:885)
- should check less than or equal to *** FAILED ***
  Error did not equal Success (CheckTest.scala:885)
```

Also note that a new test case for the >, <, >=, and <= checks has been
added under `Checks for two-columned DataFrames`.

The `assertSameJson` functionality has been refactored into a single
callsite. Additionally, this test utility function now properly sorts
the previosuly not guarenteed to be sorted lists that it has to assert
equality between. This functionality is in the `FixtureSupport` trait
of the `com.amazon.deequ.utils` package. Moreover, the
`assertJsonStringsAreEqual` helper function in the
`ConstraintSuggestionResultTest` has been updated to use this sorted
comparision fix.

#### Fix
Support for sanitizing column names for Spark SQL has been added in
the new `com.amazon.deequ.schema.ColumnName` object. The `sanitizeForSql`
function will either sanitize the column name or result in a descriptive
error. Since the rest of the deequ codebase doesn't handle errors in
a first-class way, the `sanitize` method has been added, which will
throw the error if it fails to sanitize.

The fix for ensuring that all column names that are used in Spark SQL
statements involves ensuring that all calls that would put the bare
column name into a Spark SQL statement do so only after calling
`ColumnName.sanitize`, which will wrap the name in backticks ("``").
Additionally, this fix includes understanding when it is crucial to
include the escaped column name vs. the real (unescaped) name.

#### License

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Compliance.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Compliance.scala
@@ -18,7 +18,7 @@ package com.amazon.deequ.analyzers
 
 import org.apache.spark.sql.types.IntegerType
 import org.apache.spark.sql.{Column, Row}
-import org.apache.spark.sql.functions._
+import org.apache.spark.sql.functions.{ sum, expr }
 import Analyzers._
 
 /**
@@ -31,7 +31,7 @@ import Analyzers._
   *                         Also the constraint given here can be referring to multiple columns,
   *                         so metric instance name should be provided,
   *                         describing what the analysis being done for.
-  * @param predicate SQL-predicate to apply per row
+  * @param predicate SQL-predicate to apply per row (MUST be valid Spark SQL syntax)
   * @param where Additional filter to apply before the analyzer is run.
   */
 case class Compliance(instance: String, predicate: String, where: Option[String] = None)
--- a/src/main/scala/com/amazon/deequ/analyzers/Compliance.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Compliance.scala
@@ -18,7 +18,7 @@ package com.amazon.deequ.analyzers
 
 import org.apache.spark.sql.types.IntegerType
 import org.apache.spark.sql.{Column, Row}
-import org.apache.spark.sql.functions.{ sum, expr }
+import org.apache.spark.sql.functions.{sum, expr}
 import Analyzers._
 
 /**
--- a/src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
@@ -56,9 +56,11 @@ object FrequencyBasedAnalyzer {
       numRows: Option[Long] = None)
     : FrequenciesAndNumRows = {
 
-    val columnsToGroupBy = groupingColumns.map { unsafeColumnName =>
-      col(ColumnName.sanitize(unsafeColumnName))
-    }.toArray
+    val columnsToGroupBy = groupingColumns
+      .map { unsafeColumnName =>
+        col(ColumnName.sanitize(unsafeColumnName))
+      }
+      .toArray
     val projectionColumns = columnsToGroupBy :+ col(COUNT_COL)
 
     val noGroupingColumnIsNull = groupingColumns
--- a/src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
@@ -16,6 +16,7 @@
 
 package com.amazon.deequ.analyzers
 
+import com.amazon.deequ.schema.ColumnName
 import org.apache.spark.sql.{Column, DataFrame, Row}
 import org.apache.spark.sql.functions.{coalesce, col, count, expr, lit}
 import Analyzers.COUNT_COL
@@ -23,7 +24,6 @@ import com.amazon.deequ.metrics.DoubleMetric
 import Analyzers._
 import org.apache.spark.sql.types.StructType
 import Preconditions._
-import com.amazon.deequ.analyzers.runners.MetricCalculationException
 
 /** Base class for all analyzers that operate the frequencies of groups in the data */
 abstract class FrequencyBasedAnalyzer(columnsToGroupOn: Seq[String])
@@ -56,13 +56,17 @@ object FrequencyBasedAnalyzer {
       numRows: Option[Long] = None)
     : FrequenciesAndNumRows = {
 
-    val columnsToGroupBy = groupingColumns.map { name => col(name) }.toArray
+    val columnsToGroupBy = groupingColumns.map { unsafeColumnName =>
+      col(ColumnName.sanitize(unsafeColumnName))
+    }.toArray
     val projectionColumns = columnsToGroupBy :+ col(COUNT_COL)
 
     val noGroupingColumnIsNull = groupingColumns
-      .foldLeft(expr(true.toString)) { case (condition, name) =>
-        condition.and(col(name).isNotNull)
+      .foldLeft(expr(true.toString)) { case (condition, unsafeColumnName) =>
+        condition.and(col(ColumnName.sanitize(unsafeColumnName)).isNotNull)
       }
+
+
 
     val frequencies = data
       .select(columnsToGroupBy: _*)
@@ -132,12 +136,18 @@ case class FrequenciesAndNumRows(frequencies: DataFrame, numRows: Long)
       .filterNot { _ == COUNT_COL }
 
     val projectionAfterMerge =
-      columns.map { column => coalesce(col(s"this.$column"), col(s"other.$column")).as(column) } ++
+      columns.map { unsanitizedColumn =>
+        val column = ColumnName.sanitize(unsanitizedColumn)
+        // `.as(...)` will properly escape its input column name alias
+        coalesce(col(s"this.$column"), col(s"other.$column")).as(unsanitizedColumn)
+      } ++
         Seq((zeroIfNull(s"this.$COUNT_COL") + zeroIfNull(s"other.$COUNT_COL")).as(COUNT_COL))
 
     /* Null-safe join condition over equality on grouping columns */
     val joinCondition = columns.tail
-      .foldLeft(nullSafeEq(columns.head)) { case (expr, column) => expr.and(nullSafeEq(column)) }
+      .foldLeft(nullSafeEq(ColumnName.sanitize(columns.head))) {
+        case (expr, column) => expr.and(nullSafeEq(ColumnName.sanitize(column)))
+      }
 
     /* Null-safe outer join to merge histograms */
     val frequenciesSum = frequencies.alias("this")
@@ -147,10 +157,12 @@ case class FrequenciesAndNumRows(frequencies: DataFrame, numRows: Long)
     FrequenciesAndNumRows(frequenciesSum, numRows + other.numRows)
   }
 
+  /* NOTE: Caller is responsible for ensuring that `column` is sanitized & safe for Spark SQL. */
   private[analyzers] def nullSafeEq(column: String): Column = {
     col(s"this.$column") <=> col(s"other.$column")
   }
 
+  /* NOTE: Caller is responsible for ensuring that `column` is sanitized & safe for Spark SQL. */
   private[analyzers] def zeroIfNull(column: String): Column = {
     coalesce(col(column), lit(0))
   }
--- a/src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
@@ -46,9 +46,8 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] = 
 
   override def aggregationFunctions(): Seq[Column] = {
     val sanitizedColumn = col(ColumnName.sanitize(column))
-    val expression = when(
-        regexp_extract(sanitizedColumn, pattern.toString(), 0) =!= lit(""), 1
-    ).otherwise(0)
+    val expression = when(regexp_extract(sanitizedColumn, pattern.toString(), 0) =!= lit(""), 1)
+      .otherwise(0)
 
     val summation = sum(conditionalSelection(expression, where).cast(IntegerType))
 
--- a/src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
@@ -17,6 +17,7 @@
 package com.amazon.deequ.analyzers
 
 import com.amazon.deequ.analyzers.Analyzers._
+import com.amazon.deequ.schema.ColumnName
 import org.apache.spark.sql.{Column, Row}
 import org.apache.spark.sql.functions.{col, lit, regexp_extract, sum, when}
 import org.apache.spark.sql.types.IntegerType
@@ -44,9 +45,9 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] = 
   }
 
   override def aggregationFunctions(): Seq[Column] = {
-
-    val expression = when(regexp_extract(col(column), pattern.toString(), 0) =!= lit(""), 1)
-      .otherwise(0)
+    val expression = when(
+        regexp_extract(col(ColumnName.sanitize(column)), pattern.toString(), 0) =!= lit(""), 1
+    ).otherwise(0)
 
     val summation = sum(conditionalSelection(expression, where).cast(IntegerType))
 
--- a/src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
@@ -45,8 +45,9 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] = 
   }
 
   override def aggregationFunctions(): Seq[Column] = {
+    val sanitizedColumn = col(ColumnName.sanitize(column))
     val expression = when(
-        regexp_extract(col(ColumnName.sanitize(column)), pattern.toString(), 0) =!= lit(""), 1
+        regexp_extract(sanitizedColumn, pattern.toString(), 0) =!= lit(""), 1
     ).otherwise(0)
 
     val summation = sum(conditionalSelection(expression, where).cast(IntegerType))
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -25,6 +25,8 @@ import com.amazon.deequ.metrics.{Distribution, Metric}
 import com.amazon.deequ.repository.MetricsRepository
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import com.amazon.deequ.anomalydetection.HistoryUtils
+import com.amazon.deequ.schema.ColumnName
+
 import scala.util.matching.Regex
 
 object CheckLevel extends Enumeration {
@@ -672,8 +674,9 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
+    val c = ColumnName.sanitize(column)
     // coalescing column to not count NULL values as non-compliant
-    satisfies(s"COALESCE($column, 0.0) >= 0", s"$column is non-negative", hint = hint)
+    satisfies(s"COALESCE($c, 0.0) >= 0", s"$column is non-negative", hint = hint)
   }
 
   /**
@@ -683,8 +686,9 @@ case class Check(
     * @return
     */
   def isPositive(column: String): CheckWithLastConstraintFilterable = {
+    val c = ColumnName.sanitize(column)
     // coalescing column to not count NULL values as non-compliant
-    satisfies(s"COALESCE($column, 1.0) > 0", s"$column is positive")
+    satisfies(s"COALESCE($c, 1.0) > 0", s"$column is positive")
   }
 
   /**
@@ -702,8 +706,10 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
-    satisfies(s"$columnA < $columnB", s"$columnA is less than $columnB",
-      hint = hint)
+    val (cA, cB) = ColumnName.getOrThrow(
+      (ColumnName.sanitizeForSql(columnA), ColumnName.sanitizeForSql(columnB))
+    )
+    satisfies(s"$cA < $cB", s"$columnA is less than $columnB", hint = hint)
   }
 
   /**
@@ -720,8 +726,10 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
-    satisfies(s"$columnA <= $columnB", s"$columnA is less than or equal to $columnB",
-      hint = hint)
+    val (cA, cB) = ColumnName.getOrThrow(
+      (ColumnName.sanitizeForSql(columnA), ColumnName.sanitizeForSql(columnB))
+    )
+    satisfies(s"$cA <= $cB", s"$columnA is less than or equal to $columnB", hint = hint)
   }
 
   /**
@@ -738,8 +746,10 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
-    satisfies(s"$columnA > $columnB", s"$columnA is greater than $columnB",
-      hint = hint)
+    val (cA, cB) = ColumnName.getOrThrow(
+      (ColumnName.sanitizeForSql(columnA), ColumnName.sanitizeForSql(columnB))
+    )
+    satisfies(s"$cA > $cB", s"$columnA is greater than $columnB", hint = hint)
   }
 
   /**
@@ -757,8 +767,10 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
-    satisfies(s"$columnA >= $columnB", s"$columnA is greater than or equal to $columnB",
-      hint = hint)
+    val (cA, cB) = ColumnName.getOrThrow(
+      (ColumnName.sanitizeForSql(columnA), ColumnName.sanitizeForSql(columnB))
+    )
+    satisfies(s"$cA >= $cB", s"$columnA is greater than or equal to $columnB", hint = hint)
   }
 
   // We can't use default values here as you can't combine default values and overloading in Scala
@@ -832,12 +844,13 @@ case class Check(
       hint: Option[String])
     : CheckWithLastConstraintFilterable = {
 
+    val c = ColumnName.sanitize(column)
 
     val valueList = allowedValues
       .map { _.replaceAll("'", "''") }
       .mkString("'", "','", "'")
 
-    val predicate = s"`$column` IS NULL OR `$column` IN ($valueList)"
+    val predicate = s"$c IS NULL OR $c IN ($valueList)"
     satisfies(predicate, s"$column contained in ${allowedValues.mkString(",")}", assertion, hint)
   }
 
@@ -861,11 +874,13 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
+    val c = ColumnName.sanitize(column)
+
     val leftOperand = if (includeLowerBound) ">=" else ">"
     val rightOperand = if (includeUpperBound) "<=" else "<"
 
-    val predicate = s"`$column` IS NULL OR " +
-      s"(`$column` $leftOperand $lowerBound AND `$column` $rightOperand $upperBound)"
+    val predicate =
+      s"$c IS NULL OR ($c $leftOperand $lowerBound AND $c $rightOperand $upperBound)"
 
     satisfies(predicate, s"$column between $lowerBound and $upperBound", hint = hint)
   }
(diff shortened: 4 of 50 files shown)


malcolmgreaves:
I also added myself as a developer since I've pushed a few PRs now :D

src/main/scala/com/amazon/deequ/analyzers/Compliance.scala
```diff
@@ -18,7 +18,7 @@ package com.amazon.deequ.analyzers
 
 import org.apache.spark.sql.types.IntegerType
 import org.apache.spark.sql.{Column, Row}
-import org.apache.spark.sql.functions._
+import org.apache.spark.sql.functions.{ sum, expr }
```
sscdotopen: no space after { and before } please

src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
```diff
@@ -56,14 +56,18 @@ object FrequencyBasedAnalyzer {
       numRows: Option[Long] = None)
     : FrequenciesAndNumRows = {
 
-    val columnsToGroupBy = groupingColumns.map { name => col(name) }.toArray
+    val columnsToGroupBy = groupingColumns.map { unsafeColumnName =>
+      col(ColumnName.sanitize(unsafeColumnName))
+    }.toArray
```
sscdotopen: please reformat as follows:

```
    val columnsToGroupBy = groupingColumns
      .map { unsafeColumnName =>
        col(ColumnName.sanitize(unsafeColumnName))
      }
      .toArray
```

src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
```diff
@@ -44,9 +45,9 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] =
   }
 
   override def aggregationFunctions(): Seq[Column] = {
-
-    val expression = when(regexp_extract(col(column), pattern.toString(), 0) =!= lit(""), 1)
-      .otherwise(0)
+    val expression = when(
+        regexp_extract(col(ColumnName.sanitize(column)), pattern.toString(), 0) =!= lit(""), 1
+    ).otherwise(0)
```
sscdotopen: please put ColumnName.sanitize(column) into a temporary variable as this expression is becoming to complex otherwise

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -672,8 +674,9 @@ case class Check(
       hint: Option[String] = None)
     : CheckWithLastConstraintFilterable = {
 
+    val c = ColumnName.sanitize(column)
```
sscdotopen: please don't use single char variable names, call sanitizedColumn or so

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -832,12 +844,13 @@ case class Check(
       hint: Option[String])
     : CheckWithLastConstraintFilterable = {
 
+    val c = ColumnName.sanitize(column)
```
sscdotopen: again please give the variable an easy to understand name

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,109 @@
...
+    */
+  def sanitizeForSql(columnName: String): Sanitized =
+    if (columnName == null) {
+      Left(NullColumn)
+
+    } else {
+      val (prefix, suffix, insideColumnName) = {
+        val prefix = if (!columnName.startsWith("`")) "`" else ""
+        val suffix = if (!columnName.endsWith("`")) "`" else ""
+        val inside1 = if (prefix.isEmpty) {
```
sscdotopen: please another name to this variable to make the code easier to read

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,109 @@
...
+          columnName.slice(1, columnName.length)
+        } else {
+          columnName
+        }
+        val inside2 = if (suffix.isEmpty) {
```
sscdotopen: please give another name to this variable

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,109 @@
...
+    case (Right(cA), Right(cB)) => (cA, cB)
+    case (Left(eA), Left(eB)) => throw new IllegalArgumentException(
+      s"Cannot sanitize two column names:\n$eA\n$eB"
+    )
+    case (Left(e), _) => throw e
+    case (_, Left(e)) => throw e
+  }
+
+  /** Alias for `sanitizeForSql | getOrThrow`. */
+  def sanitize(columnName: String): String =
```
sscdotopen: please add brackets here

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+    * Obtains the `String` pair if both are `Right`, otherwise throws the error(s).
+    *
+    * If only one of the sanitizations failed, then a `SanitizeError` type is thrown.
+    * If both fail, then an `IllegalArgumentException` is thrown and its message contains
+    * both of the `SanitizeError` messages.
+    */
+  def getOrThrow(x: (Sanitized, Sanitized)): (String, String) = x match {
+    case (Right(cA), Right(cB)) => (cA, cB)
```
sscdotopen: please use variable names that are easier to understand

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,109 @@
...
+    case (Left(e), _) => throw e
+    case (_, Left(e)) => throw e
+  }
+
+  /** Alias for `sanitizeForSql | getOrThrow`. */
+  def sanitize(columnName: String): String =
+    getOrThrow(sanitizeForSql(columnName))
+
+  /** Inverse of `sanitize`: removes surrounding backticks, if present. */
+  def desanitize(maybeSanitizedName: String): String =
```
sscdotopen: why cant we use the replaceAll method with ` as argument?

src/main/scala/com/amazon/deequ/suggestions/rules/CategoricalRangeRule.scala
```diff
@@ -45,6 +46,8 @@ case class CategoricalRangeRule() extends ConstraintRule[ColumnProfile] {
 
   override def candidate(profile: ColumnProfile, numRecords: Long): ConstraintSuggestion = {
 
+    val c = ColumnName.sanitize(profile.column)
```
sscdotopen: please use more meaningful variable names

src/main/scala/com/amazon/deequ/suggestions/rules/FractionalCategoricalRangeRule.scala
```diff
@@ -52,6 +54,8 @@ case class FractionalCategoricalRangeRule(targetDataCoverageFraction: Double = 0
 
   override def candidate(profile: ColumnProfile, numRecords: Long): ConstraintSuggestion = {
 
+    val c = ColumnName.sanitize(profile.column)
```
sscdotopen: please use more meaningful variable names

sscdotopen:
On 13.02.19 20:13, Malcolm Greaves wrote:
> I also added myself as a developer since I've pushed a few PRs now :D

Sure, welcome :)

src/test/scala/com/amazon/deequ/checks/CheckTest.scala
```diff
@@ -801,8 +877,20 @@ object CheckTest extends WordSpec with Matchers {
       dataType, sparkSession)
   }
 
-  val badColumnName: String = "[this column]:has a handful of problematic chars"
+  def testCheckOnData(df: DataFrame, c: Check): Unit = {
+    val r = VerificationSuite()
```
sscdotopen: please give a better name to r

src/test/scala/com/amazon/deequ/schema/ColumnNameTest.scala
```diff
@@ -0,0 +1,119 @@
...
+package com.amazon.deequ.schema
+
+import org.scalatest.{Matchers, WordSpec}
+
+class ColumnNameTest extends WordSpec with Matchers {
+
+  "ColumnName's sanitizeForSql function" should {
+
+    "leave escaped column name unchanged" in {
+      val c = "`escaped_column_name`"
```
sscdotopen: please give a better name to c

src/test/scala/com/amazon/deequ/utils/FixtureSupport.scala
```diff
@@ -202,21 +205,53 @@ trait FixtureSupport {
...
 
     import sparkSession.implicits._
     (1 to numberOfRows)
       .toList
       .map { index => (s"$index", random.shuffle(categories).head)}
-      .toDF("att1", "categoricalColumn")
+      .toDF("]att1[", "categoricalColumn")
+  }
+
+  implicit object OrderingTestMap extends Ordering[Map[String, Any]] {
```
sscdotopen: can you explain why we need this?

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: We could do that, especially since `maybeSanitizedName` _should_ only have backticks in the beginning and end, so removing all `\`` _should_ be equivalent in our use cases.

I did, however, write this function with the explicit intent of being an inverse of `sanitize`. I can def change the implementation to `_.removeAll("`")`. Given this intent, for the review, should I change?

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: Sure thing! These are successfully sanitized column values. I originally didn't want to keep this function def that long since it's merely mechanically unpacking the sanitized values (if present) from their error-aware wrapping. I'll follow a convention I am using from this review -- renaming such uses as `sanitizedColumn`.

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: Re-wrote this awkward definition to slice into the string directly.

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: Re-wrote this awkward definition to slice into the string directly. I think this new definition is much more straightforward _and_ it is an inverse of `sanitize`:
```scala
maybeSanitizedName.slice(
    if(maybeSanitizedName.startsWith("`")) 1 else 0,
    maybeSanitizedName.length - {if(maybeSanitizedName.endsWith("`")) 1 else 0 }
)
```

src/test/scala/com/amazon/deequ/utils/FixtureSupport.scala
malcolmgreaves: I added this comment:

(tl;dr to make the JSON deserialize tests ignore ordering since most of the tests don't need to explicitly check for ordering)

```
A map ordering to ensure that JSON array deserialization results are
checked in a consistent manner.

In general, an `Ordering[Map[String,Any]]` is not a great pattern: here,
however, we have a need to check that various JSON deserialization-based
tests correctly parse & handle arrays of objects. Instead of ensuring
that the ordering inherint in the element processing of the various bits
of code under test do not permute their outputs (something that _is not_
a part of any of the code's stated contracts), the idea here is to re-use
_this_ ordering definition.

This JSON map ordering definiton is specific to the deequ tests: the calls
insde the over-loaded `compare` method have a specific, deterministic
order of hard-coded keys to check when determining which of two JSON
deserialized objects is before another in a sorted collection.
```

malcolmgreaves:
👋 @sscdotopen I have addressed all review comments. Please let me know if I missed something or if you have any new comments for review.

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: Unfortunately, I had to roll-back this change (broke tests :-/). I updated names to be consistent w/ the ones in `sanitizeForSql`.

src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
sscdotopen: I think we should break this up into a few more statements, still looks to complicated :)

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,110 @@
...
+    * The resulting String is the escaped input column name, which is safe to use in
+    * any Spark SQL statement.
+    */
+  def sanitizeForSql(columnName: String): Sanitized =
+    if (columnName == null) {
+      Left(NullColumn)
+
+    } else {
+      val (prefix, suffix, rawColumnName) = {
+        val prefix = if (!columnName.startsWith("`")) "`" else ""
```
sscdotopen: I think this can be simplified, because columnName.startsWith("`") in this line and prefix.isEmpty later basically check the same condition

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,110 @@
...
+    getOrThrow(sanitizeForSql(columnName))
+  }
+
+  /** Inverse of `sanitize`: removes surrounding backticks, if present. */
+  def desanitize(maybeSanitizedName: String): String =
+    if (maybeSanitizedName == null) {
+      ""
+    } else {
+      val nameWoPrefix =
+        if (maybeSanitizedName.startsWith("`")) {
```
sscdotopen: can we refactor the code that checks the last char into a method and use that method instead of copying the code?

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: It's also used in creating the sanitized string (in the final `if-else` expression: the `Right(s"$prefix$rawColumnName$suffix")` expression.

malcolmgreaves: Pushed a change that I think simplifies this all:
- raw column name extraction is a helper function
- `startsWith` & `endsWith` checks are in helper functions

```scala
      val rawColumnName = extractColumnName(columnName)
      if (rawColumnName.contains("`")) {
        Left(ColumnNameHasBackticks(columnName))
      } else {
        val prefix = if (hasStartingTick(columnName)) "" else "`"
        val suffix = if (hasEndingTick(columnName)) "" else "`"
        Right(s"$prefix$rawColumnName$suffix")
      }
```

malcolmgreaves:
```scala

  /** Extracts the original column name from a possibly sanitized column name. */
  @inline
  private[this] def extractColumnName(name: String): String =
    name.slice(
      if (hasStartingTick(name)) 1 else 0,
      name.length - (if (hasEndingTick(name)) 1 else 0)
    )

  /** True iff the input string starts with a backtick. False otherwise. */
  @inline
  private[this] def hasStartingTick(s: String): Boolean =
    s.startsWith("`")

  /** True iff the input string ends with a backtick. False otherwise. */
  @inline
  private[this] def hasEndingTick(s: String): Boolean =
    s.endsWith("`")
```

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: Yup! :D

```
  /** True iff the input string starts with a backtick. False otherwise. */
  @inline
  private[this] def hasStartingTick(s: String): Boolean =
    s.startsWith("`")

  /** True iff the input string ends with a backtick. False otherwise. */
  @inline
  private[this] def hasEndingTick(s: String): Boolean =
    s.endsWith("`")
```

malcolmgreaves:
@sscdotopen Simplified `sanitizeForSql` & `desanitize` defs by refactoring common-code into helper functions. Also re-worked the "raw" column name extraction to be a simple string slice operation.

For `PatternMatch.scala`: I did re-work the code to be the _nearly_ the same form as the original. It was:
```scala
  override def aggregationFunctions(): Seq[Column] = {
    val expression = when(regexp_extract(col(column), pattern.toString(), 0) =!= lit(""), 1)
...
```
And it now is:
```scala
  override def aggregationFunctions(): Seq[Column] = {
    val sanitizedColumn = col(ColumnName.sanitize(column))
    val expression = when(regexp_extract(sanitizedColumn, pattern.toString(), 0) =!= lit(""), 1)
      .otherwise(0)
...
```

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+    * Obtains the `String` pair if both are `Right`, otherwise throws the error(s).
+    *
+    * If only one of the sanitizations failed, then a `SanitizeError` type is thrown.
+    * If both fail, then an `IllegalArgumentException` is thrown and its message contains
+    * both of the `SanitizeError` messages.
+    */
+  def getOrThrow(x: (Sanitized, Sanitized)): (String, String) = x match {
```
sscdotopen: where exactly is this used? I'm not so happy with a method that could throw an exception from some place in the code.

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,100 @@
...
+  @inline
+  private[this] def extractColumnName(name: String): String =
+    name.slice(
+      if (hasStartingTick(name)) 1 else 0,
+      name.length - (if (hasEndingTick(name)) 1 else 0)
+    )
+
+  /** True iff the input string starts with a backtick. False otherwise. */
+  @inline
+  private[this] def hasStartingTick(s: String): Boolean =
```
sscdotopen: please use brackets here {}

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
```diff
@@ -0,0 +1,100 @@
...
+    s.startsWith("`")
+
+  /** True iff the input string ends with a backtick. False otherwise. */
+  @inline
+  private[this] def hasEndingTick(s: String): Boolean =
```
sscdotopen: please use brackets here {}

sscdotopen:
I have a quick question: where exactly would this code throw exceptions? It would be great if we could check the validity of the column names in Analyzer.preconditions https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala#L76 and throw an exception immediately when there is a problem.

src/main/scala/com/amazon/deequ/schema/ColumnName.scala
malcolmgreaves: This is used in `sanitize`, which is used all-over the place: whenever there's a column name that is going into a Spark SQL expression. I would prefer to not throw exceptions too. However, the code written as-is assumes that columns would always be sanitized. As a result, the various Spark SQL-calling code doesn't have return types that encode an error: they always assume that the computation will succeed.

As it currently stands, `deequ`'s current behavior is to already throw an exception when using an unsanitzied column name with special characters during Spark SQL execution: it would be a Spark SQL `SyntaxError`. So this PR doesn't stray too far from existing functionality. In fact, it would only throw an exception now if the column name could not be sanitized (i.e. has backticks  in the name -- not a pair in the surrounding/escaped position).

malcolmgreaves:
@sscdotopen for where, see https://github.com/awslabs/deequ/pull/91#discussion_r260541274

I am not sure how that method signature checks column problems: it's a sequence of side-effecting functions (`StructType => Unit`). How do they return useful information about their assertions? Shouldn't they be `StructType => Option[Exception]` ? Or something similar? Or, do you propose to throw an exception? So not throwing an exception for one of these "precondition" functions implies that it is good?

Even if this would occur, the problem is still that we have to sanitize the column names _at every place_ one would go into a Spark SQL statement.

I did consider an alternative design: having a tagged type (`String` implementation) for a sanitized vs. unsanitzed column name. This would have the benefit of having a single process point for converting a column name into a sanitized one -- handling errors at that point in time -- and then eliminating the exception-throwing behavior in the various `check` calls. Additionally, it wouldn't have any runtime overhead as it would be a tagged type (e.g. http://eed3si9n.com/learning-scalaz/Tagged+type.html).

However, due to some inconsistencies w/ Spark's SQL API, sometimes having backticks in the input column name is a problem (I forget the exact ones, but they're in this PR as things that are calling into the API and explicitly _don't_ call `ColumnName.sanitize` bc. Spark will already wrap the column name in backticks :-/). So, even if we did this, we would have to be a little careful. It's unfortunately not as straightforward that a sanitized column name _always_ works.

malcolmgreaves:
After looking at `preconditions` and the `Analyzer` type, I realize that there are no columns listed as a member of `Analyzer`. So, doing this column name sanitization would have to happen on a per-implementing type basis. As a result, I don't think `preconditions` is the best spot here: it seems to me that the point of `preconditions` is to allow an implementing class to provide arbitrary, custom checks. What we want is to enforce a single type of precondition for all implementations.

If we want to go-ahead with this change to `Analyzer`, we could (1) re-work `preconditions` to have a over-rideable option and also include the column name sanitization or (2) identify a spot before any `Analzyer` is run to check for column name sanitization (and not change `Analzyer`).

For (1): we could (a) add a `def columns(): Seq[StructType]` method to `analyzer`, (b) refactor `def preconditions` to `protected def custom_preconditions`, and (c) define

```scala
final val preconditions: Seq[StructType => Unit] = 
    columns().map { c => assert ColumnName.santizieForSql(c.name).isRight } ++ custom_preconditions()
```

For (2): I would need some help in identifying such a spot :D

sscdotopen:
Hi Malcolm,

I thought a little bit about the general direction of this (and about the chance of exceptions being thrown). I think we can simplify this quite a bit:

 * Rather than implementing the name escaping/sanitizing ourselves, we should rely on Spark's [Column](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/Column.scala) class (used by the **col** function), which should offer a way for us to do this. 

 * We should change the code where we create SQL or SQL snippets using strings to leverage Spark's programmatic SQL API

 * This should only leave us with a few places where exceptions could be thrown and (hopefully) thats only the places where the user supplies SQL snippets that we have to trust.

 What do you think about that direction?

malcolmgreaves:
@sscdotopen Are you proposing a public API refactor? From what I understand of your proposal, using a Spark `Column` as a parameter type for column names (instead of `String`) would mean that all of the methods in `Check.scala` and a few other ones (mainly analyzer-related I believe) would have their public members signatures change. Is my understanding correct here?

From what I read, creating an instance [1] for a name that needs sanitization (e.g. a column named `]att1[`, which is a part of this PR's updated tests) a would hit [2]. They have a method [3] that parses the name according to rules: `\`` are in pairs, globs `*`  are ok, and path parts `.` must have characters between 'em.

However, If parsing the name fails, then this code throws an exception [4]: `new AnalysisException(s"syntax error in attribute name: $name")`. It does not attempt to sanitize it and continue if successful.



**If we refactor the API** to use `Column` instead of `String` when appropriate, then we can be sure that we're always dealing with sanitized values. If we go this route, then:
1. we will have breaking changes (ok to do I suppose since we're pre-1.0!)
2. I suggest we refactor `ColumnName.sanitizeForSql` to be `String => Either[SanitizeError, Column]` and remove it's uses from w/in the analyzer and check code

For (2), I am suggesting we keep around a `deequ`-blessed way to go from a user's column name to one that is sanitized. We can keep the PR's existing code that attempts to sanitize a column name that, w/o surrounding backticks, _would_ cause an `AnalysisException` from being thrown upon a `new Column(...)` call. Additionally, I would update `SanitizeError` with a new case that would be a catch-all for an `AnalysisException`. In psuedo-Scala, this would be something such as:
```scala

case class FailColumn(error: exception) extends SanitizeError
case class ColumnParsingError(error: AnalysisException) extends FailColumn(error)

def sanizieForSql(columnName: String): Either[SanitizeError, Column] = {
    old_sanitizeForSql(columnName).flatMap { sanitized =>
          Try(new Column(sanitized)) match {
              case Failure(e) => e match {
                  case ae: AnalysisException => ColumnParsingError(ae)
                  case _ => FailColumn(e)
              }
              case Success(c) => c
          }
      }
}
```
(some details are missing ofc: backtick-checking in "old" impl would be removed as that's done in `new Column`)



If, in fact, **we don't_ want to change the API**, then I think we would want to only add the `AnalysisException`-throwing call to `spark.sql.catalyst.analysis.unresolved.parseAttributeName` as an additional check in `ColumnName.sanitizeForSql`. `parseAttributeName` does additional name format checks w/ `*` and `.` and it also would not fail on more complicated, yet valid, column names with non-surrounding backticks (learned from that impl. that they need only come in pairs of backticks).

------------------------------------------------------------------------------------------------
[1] https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/Column.scala#L141
[2] https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/unresolved.scala#L142
[3] https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/unresolved.scala#L157
[4] https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/unresolved.scala#L180

sscdotopen:
Hi Malcolm,

I'm not proposing to change the public API. I think we should discuss two things:

 * We should not have to implement column name sanitation ourselves, but rather use whatever is provided by Spark.

 * I think a main problem is that we build SQL queries (or parts of them) using string concatenation. We should not change the public API, but rather internally build the queries or parts of them using Spark's programmatic functions like col, expr and so on. My gut feeling is that this will automatically solve the column name sanitation problem as well.

Let me know what you think about that.

malcolmgreaves:
@sscdotopen Yes, wonderful! I appreciate the clarification. That makes a lot of sense: we're serializing _for_ SparkSQL, so we should use their method of column name serialization. Not using this originally was a complete oversight on my part. I will change the PR to use the serialization provided by `Column`. So, `ColumnName` will go (or be transformed at the very least).

For the second part of the new fix, building Spark SQL queries sans string concatenation: this sounds like a great way to build some flexible structure for the whole project. I agree that it will most likely reduce the presence of bugs: we should ensure that, w/in a query representation, all column names have been sanitized.

What are your thoughts on the design of such a query-building data structure. Would an accompanying small DSL be appropriate as well? Perhaps it could be closely modeled from the existing ways `deequ` expresses data constraints? Or, maybe we could have an ADT-like system: a type that represents a complete, sanitized, ready-to-execute query; another that represents a sanitized, yet incomplete query (needs more terms to be satisfied); and a base type that is a sanitized column name.

We need to do error handling on the base-case creating step: ensuring a column input (`String`) is sanitized using Spark's `Column` constructor means we'll throw an exception.

I think that this work would be fairly involved as it would underlie many parts of the library. I think it's best to try to work out some design goals and requirements before diving into writing code. I want to make sure I write the _right_ thing :D

sscdotopen:
Hi Malcolm, I think that Spark also already has a kind of DSL for safe query building that we should leverage. Details can be found in [org.apache.spark.sql.functions._](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/functions.scala).

However, if you would be interested in contributing something larger, more involved to deequ, than I'm sure we would find a project for you that you could enjoy. Let me know if you'd be interested in that, and we could have a skype call on that topic.

Best,
Sebastian

malcolmgreaves:
Hi @sscdotopen -- it's been awhile! I have been busy with other projects at work and some other things in life! And I haven't found time to commit yet to such a large undertaking -- re-writing the entire PR with a brand new query-building system.

Good news is that I might have some more time coming up in the next few weeks to dedicate to working on `deequ` again. I would like to sync up and discuss design goals & objectives with you on this bug fix!

sscdotopen:
I am closing this due to inactivity. Feel free to reopen the PR later.

PR #94, opened 2019-03-15
Adapts AttributeReference call to changes made in Spark 2.4

iamsteps:
*Description of changes:*

In Spark 2.4 [[Spark-19602]](https://github.com/apache/spark/commit/b4bf8be549afa51e931e48dd79ddd9480f567b13) the signature for `AttributeReference` has changed (`qualifier` argument changed from `Option[String]` to `Seq[String]`), causing an `IllegalArgumentException` for some of our analyzers (`ApproxCountDistinct`) when used in conjunction with Spark 2.4. This PR adds a condition on the argument type (via reflection) for `AttributeReference`, in order to distinguish behavior for Spark 2.4 from previous versions.

Successfully tested against Spark 2.2, Spark 2.3, and Spark 2.4.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
+++ b/src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
@@ -100,14 +100,17 @@ public class AttributeReferenceCreation {
                         emptyMetadata, exprId, none, false);
             } else {
                 // Spark 2.4
-                if (Seq.class.isAssignableFrom(apply.getParameterTypes()[5])) {
+                Class<?> qualifierParameterType = apply.getParameterTypes()[5];
+                boolean qualifierParameterTypeIsSeq =
+                    Seq.class.isAssignableFrom(qualifierParameterType);
+                if (qualifierParameterTypeIsSeq) {
                     return (AttributeReference) apply.invoke(companion, name, dataType, true,
                             emptyMetadata, exprId, Seq$.MODULE$.<String>empty());
+                } else {
+                    // Spark 2.3
+                    return (AttributeReference) apply.invoke(companion, name, dataType, true,
+                            emptyMetadata, exprId, none);
                 }
-
-                // Spark 2.3
-                return (AttributeReference) apply.invoke(companion, name, dataType, true,
-                        emptyMetadata, exprId, none);
             }
         } catch (Exception e) {
             throw new IllegalStateException(e);
--- a/src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
+++ b/src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
@@ -26,6 +26,8 @@ import org.apache.spark.sql.types.Metadata;
 import org.apache.spark.sql.types.Metadata$;
 
 import java.lang.reflect.Method;
+import scala.collection.Seq;
+import scala.collection.Seq$;
 
 public class AttributeReferenceCreation {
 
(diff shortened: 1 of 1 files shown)


src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
```diff
@@ -87,6 +99,12 @@ public static AttributeReference createSafe(String name) throws IllegalStateExce
                 return (AttributeReference) apply.invoke(companion, name, dataType, true,
                         emptyMetadata, exprId, none, false);
             } else {
+                // Spark 2.4
+                if (Seq.class.isAssignableFrom(apply.getParameterTypes()[5])) {
+                    return (AttributeReference) apply.invoke(companion, name, dataType, true,
+                            emptyMetadata, exprId, Seq$.MODULE$.<String>empty());
+                }
```
dustin-lange: add explicit else

src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
```diff
@@ -26,6 +26,8 @@
 import org.apache.spark.sql.types.Metadata$;
 
 import java.lang.reflect.Method;
+import scala.collection.Seq;
+import scala.collection.Seq$;
```
dustin-lange: What is Seq$? Is it for the typed Seq?

src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
```diff
@@ -87,6 +99,12 @@ public static AttributeReference createSafe(String name) throws IllegalStateExce
                 return (AttributeReference) apply.invoke(companion, name, dataType, true,
                         emptyMetadata, exprId, none, false);
             } else {
+                // Spark 2.4
+                if (Seq.class.isAssignableFrom(apply.getParameterTypes()[5])) {
```
sscdotopen: would be better to break this up in to several lines and use temporary variables. It is hard to understand whats happening here otherwise.

src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
```diff
@@ -37,6 +39,16 @@ private AttributeReferenceCreation() { }
         signature in different versions of Spark 2.x. Therefore we need to invoke it via reflection
         depending on which version of Spark we run.
 
+        SPARK 2.4:
```
tdhd: We could add more safe guards around the if checks. For example any `AttributeReference` with 7 parameters is invoked with Spark 2.2 signature. Can we rather check all of the arguments of `AttributeReference` instead?

iamsteps: There are a lot of things we can check for sure. My preference is to "let it crash" in the tests, rather than resorting to defensive programming. Maybe you can elaborate more on your view: what would you check, and - more importantly - how exactly would you deal with previously unseen signatures?

sscdotopen: I dont think we have to handle unseen signatures, we should just make sure to correctly handle all major Spark 2.x versions

tdhd: I meant to check the whole signature of all known versions instead of checking the number of arguments only.

iamsteps: That's what I understood. Personally, I don't think that the defensive approach adds much here because 1/ we discover problems when running the test suite for a new Spark release, 2/ when discovering an unexpected argument type we are already in an illegal state and there's no remedy, so the only benefit we gain is a more informative exception. With these points combined, I'm not in favor of checking the signature exhaustively.

tdhd: Good point, fully agree. A more informative exception can help in pinpointing the issue for new Spark releases.

src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java
iamsteps: The `Seq` singleton object

PR #117, opened 2019-07-24
Lift the private package restriction on analysis result and corresposing serde object

alexjbush:
*Description of changes:*

Currently there is no way to implement a custom metrics repository as the `AnalysisResult` class is private to deequ. This is the return type of `get()` in the `MetricsRepositoryMultipleResultsLoader` trait, therefore there is no way to actually implement this trait.

There is also no way to serialise this result object so it can be stored in a custom way.

This PR lifts the private restriction on both the `AnalysisResult` class and it's corresponding serialisation object.


sscdotopen:
Thank you for your contribution!

Issue #135, opened 2019-08-29
runOnAggregatedStates with saveStatesWith not persisting all metrics


romansegador:
I'm trying to persist the state after calculate metrics from already persisted states. 

Functional Scenario is:
- Each day metrics are calculated for the given day and the delta is persisted with HdfsStateProvider
- With AnalysisRunner.runOnAggregatedStates, we calculate the metrics with the states saved as the aggregated until yesterday and the new daily delta. The result is the new total aggregated for tomorrow.

On the second case, when is used the "saveStatesWith" parameter, the state is not persisted (or at least not completely). 
example:
```
val stores: Seq[HdfsStateProvider] = Seq(hdfsStateStoreDay, hdfsStateStoreAggregatedUntilYesterday)

val aggregatedMetrics = AnalysisRunner.runOnAggregatedStates(
            schema= dataFrame.schema,
            analysis = analysis,
            stateLoaders = stores,
            saveStatesWith = Some(hdfsStateStoreAggregated)
          )

```

Digging into the code I've realized that it is only stored for groupingAnalyzers but not for scanningAnalyzers

```
 /* Identify analyzers which require us to group the data */
    val (groupingAnalyzers, scanningAnalyzers) =
      passedAnalyzers.partition { _.isInstanceOf[GroupingAnalyzer[State[_], Metric[_]]] }

    val nonGroupedResults = scanningAnalyzers
      .map { _.asInstanceOf[Analyzer[State[_], Metric[_]]] }
      .flatMap { analyzer =>
        analyzer
          .loadStateAndComputeMetric(aggregatedStates)
          .map { metric => analyzer -> metric }
      }
      .toMap[Analyzer[_, Metric[_]], Metric[_]]


    val groupedResults = if (groupingAnalyzers.isEmpty) {
      AnalyzerContext.empty
    } else {
      groupingAnalyzers
        .map { _.asInstanceOf[GroupingAnalyzer[State[_], Metric[_]]] }
        .groupBy { _.groupingColumns().sorted }
        .map { case (_, analyzersForGrouping) =>

          val state = findStateForParticularGrouping(analyzersForGrouping, aggregatedStates)

          runAnalyzersForParticularGrouping(state, analyzersForGrouping, saveStatesWith,
            storageLevelOfGroupedDataForMultiplePasses)
        }
        .reduce { _ ++ _ }
    }
```

I think it is a bug and should be also saved on this cases but please let me know your feedback. 

Thanks

[...]

PR #136, opened 2019-08-30
Issue-135: persist aggregated  state for non group analyzers

romansegador:
*Issue #135*

*Description of changes:*
while running on aggregated states, added a change to persist the no grouped Analyzers if the save state store is provided*

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
@@ -404,6 +404,9 @@ object AnalysisRunner {
     passedAnalyzers.foreach { analyzer =>
       stateLoaders.foreach { stateLoader =>
         analyzer.aggregateStateTo(aggregatedStates, stateLoader, aggregatedStates)
+        if(saveStatesWith.isDefined) {
+          analyzer.persistStateTo(aggregatedStates, saveStatesWith.get)
+        }
       }
     }
 
--- a/src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
+++ b/src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
@@ -271,5 +271,37 @@ class AnalysisRunnerTests extends WordSpec with Matchers with SparkContextSpec w
         inputStream => assert(inputStream.read() > 0)
       }
     }
+
+    "should save states when running on aggregated states" in withSparkSession { session =>
+
+      val dataFrames = getDFsWithNRowsAndUniqueValuesCrossDF(session, 2, 2)
+
+      val stateStores = Seq(InMemoryStateProvider(), InMemoryStateProvider())
+
+      val analysis = Analysis().addAnalyzers(Size() :: Uniqueness("c1") :: Nil)
+
+      for((dataFrame, stateStore) <- dataFrames zip stateStores) {
+        AnalysisRunner.run(
+          data = dataFrame,
+          analysis = analysis,
+          saveStatesWith = Some(stateStore)
+        )
+      }
+
+      val mergedStateStore = InMemoryStateProvider()
+
+      AnalysisRunner.runOnAggregatedStates(
+        schema = dataFrames.head.schema,
+        analysis = analysis,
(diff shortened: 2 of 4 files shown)


src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
```diff
@@ -404,6 +404,9 @@ object AnalysisRunner {
     passedAnalyzers.foreach { analyzer =>
       stateLoaders.foreach { stateLoader =>
         analyzer.aggregateStateTo(aggregatedStates, stateLoader, aggregatedStates)
+        if(saveStatesWith.isDefined) {
```
sscdotopen: I dont think this is the correct place to persist the aggregates, as these are only the initial aggregates before the actual delta of the data is processed.

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
```diff
@@ -271,5 +271,37 @@ class AnalysisRunnerTests extends WordSpec with Matchers with SparkContextSpec w
...
+
+    "should save states when running on aggregated states" in withSparkSession { session =>
+
+      val dataFrames = getDFsWithNRowsAndUniqueValuesCrossDF(session, 2, 2)
+
+      val stateStores = Seq(InMemoryStateProvider(), InMemoryStateProvider())
+
+      val analysis = Analysis().addAnalyzers(Size() :: Uniqueness("c1") :: Nil)
+
+      for((dataFrame, stateStore) <- dataFrames zip stateStores) {
```
sscdotopen: I dont quite understand this code. What are the semantics of ```dataFrames``` here? Is it different versions of the same data? Or is a growing dataset such that the union of all the data frames makes up the dataset?

romansegador:
Hi Sebastian let's me try to explain the use case:

We want to have aggregated information for metrics from a huge amount of data. That won't allow us to query the data for the whole table and we need to store intermediate states.

For example:

- Step1: Each day metrics are calculated for the records created on the last 24h. At this point, we will have information for each day calculated and stored with  a HdfsStateProvider  (partitions per day)
- Step2: Weekly, we use this daily saved states to calculate the weekly aggregated metrics, also storing the state with a HDFSStateProvider. At this moment, we won't query for any new data, we just use the daily states.  
- Step3: We have also a complete metrics aggregated in a similar way that number 2.

Following this approach, if a reprocess of the real data happens or some data in the past gets updated, we can just reprocess the data quality metrics for that date and recalculate from that date using the different daily / weekly / etc stores the global quality metric.

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
romansegador: this is two different dataframes with two different set of data,  Then a different stateStore is generated and linked for each dataframe.

After,  metrics for each independent dataFrame are calculated and the state stored on its linked stateStore.
(It could be seen as Monday data and metrics for Monday on same for Tuesday).

Then this states are aggregated with AnalysisRunner.runOnAggregatedStates without any new data, just the previous saved states and saved on a new StateProvider. (Which would represent the Monday + Tuesday aggregated and stored).

src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
romansegador: Probably I'm probably not following properly the code  (sorry, I'm not familiarised with Scala).

After this point I can't see any modification on the aggregated states for the "scanningAnalyzers", as they only read this states but not modify them.

For "groupingAnalyzers" they are actually modified but they also override them on the "saveStatesWith" so the final result is the expected one. 

Please let me know what would be the proper place to locate it and I will move.

sscdotopen: Hi Roman,

In general your code would be correct. The problem is just that the states for the grouping analyzers can become quite large and we need to be careful when persisting them. We have an optimization where we only persist large states once if they can be reused by multiple grouping analyzers.

PR #143, opened 2019-09-17
Update README.md

groodt:
Issue awslabs#139

Description of changes:
Clarify documentation around Spark version support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Deequ is a library built on top of Apache Spark for defining "unit tests for dat
 
 ## Requirements and Installation
 
-__Deequ__ depends on Java 8 and is known to work with Apache Spark versions 2.2.x to 2.4.x. However, be aware of the risk that deequ is extending certain internal Spark APIs which are subject to change in future and may break as Spark versions change.
+__Deequ__ depends on Java 8 and is known to work with Apache Spark versions 2.2.x to 2.4.x.
 
 Available via [maven central](http://mvnrepository.com/artifact/com.amazon.deequ/deequ). 
 Add the latest release as a dependency to your project:
--- a/README.md
+++ b/README.md
@@ -8,9 +8,7 @@ Deequ is a library built on top of Apache Spark for defining "unit tests for dat
 
 ## Requirements and Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,8 +8,10 @@ Deequ is a library built on top of Apache Spark for defining "unit tests for dat
 
 ## Requirements and Installation
 
-__Deequ__ depends on Java 8 and Apache Spark 2.2, and is available via [maven central](http://mvnrepository.com/artifact/com.amazon.deequ/deequ). Add the latest release as a dependency to your project:
+__Deequ__ depends on Java 8 and Apache Spark 2.2. It only uses non-experimental APIs, which should work up to Spark 2.4.
```
sscdotopen: It is true that we support Spark 2.3 and 2.4.

However, we extend some internal catalyst-based aggregation functions and have to apply some hacks to get this to work across different versions: https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/catalyst/AttributeReferenceCreation.java#L73

So the comment about the non-experimental APIs is not true unfortunately.

groodt: Ok, thanks for clarifying. I've updated the wording a bit to make the risks clearer.

README.md
```diff
@@ -8,8 +8,10 @@ Deequ is a library built on top of Apache Spark for defining "unit tests for dat
 
 ## Requirements and Installation
 
-__Deequ__ depends on Java 8 and Apache Spark 2.2, and is available via [maven central](http://mvnrepository.com/artifact/com.amazon.deequ/deequ). Add the latest release as a dependency to your project:
+__Deequ__ depends on Java 8 and is known to work with Apache Spark versions 2.2.x to 2.4.x. However, be aware of the risk that deequ is extending certain internal Spark APIs which are subject to change in future and may break as Spark versions change.
```
sscdotopen: I think the "However ..." sentence unnecessarily scares users. Maybe we should just say that it works up to 2.4

groodt: I think that's fine too. Ok, I'll remove that entire sentence. 👍

sscdotopen:
Thank you for your contribution!

Issue #92, opened 2019-02-18
Support for state aggregation and persistence for VerificationSuite fluent API


tdhd:
The fluent API does currently not offer setting two parameters for state aggregation and state saving.

A workaround for this is to use the deprecated run method:
```
val stateProvider = InMemoryStateProvider()
VerificationSuite().run(df, checks, saveStatesWith = Some(stateProvider))
val state = stateProvider.load(analyzer)
```

Implementation of the two additional parameters should be straightforward by adding a method for aggregation and one for state saving in `VerificationRunBuilder` possibly with this signature:
```
class VerificationRunBuilder {
  // ...
  def aggregateWith(stateLoader: StateLoader): this.type
  def saveStatesWith(statePersister: StatePersister): this.type
  // ...
}
```

and passing them to `VerificationSuite().doVerificationRun()`.

PR #145, opened 2019-09-23
WIP - adds state provider support to VerificationRunBuilder (#92)

andreas-grafberger:
*Issue #, if available:* #92 

*Description of changes:*   
Adds two methods to [`VerificationRunBuilder`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L28) to allow saving and aggregation of states. 
Currently the only way to do this with the [`VerificationSuite`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationSuite.scala#L42) is by using the deprecated [`run`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationSuite.scala#L67) method.

This would allow users to use the stateful metrics computation features of deequ without relying on deprecated methods. However, I would like to use this draft PR as a place for discussion as the way this PR implements this might not be optimal for all users. Some problems that I think should be up for discussion are:
1. This PR mainly caters towards users with datasets that grow in size over time. A new partition represents the new entries since the last verification run. However, the proposed change does not make it easier to aggregate multiple states e.g. across various partitions from different regions. For this, the method [`runOnAggregatedStates`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationSuite.scala#L208) would still be the primary way to do this. My question:
Should the [`VerificationRunBuilder`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L28) offer the user to aggregate *multiple* saved states *in addition* to the data provided via [`onData(data: DataFrame)`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationSuite.scala#L49)?

2. If a user stores states with the [`StateProvider`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/analyzers/StateProvider.scala#L72) (that implements both the [`StateLoader`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/analyzers/StateProvider.scala#L36) and [`StatePersister`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/analyzers/StateProvider.scala#L41)), they have to find appropriate ways to save/load the states in a manner that allows nice separation of datasets and partitions/ dates (or other criteria) and to correctly attribute and load them later. This can be currently done via location prefixes to the stored path. As I see it, the [`MetricsRepository`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/repository/MetricsRepository.scala#L25) makes it easier for the user by relying on [`ResultKeys`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/repository/MetricsRepository.scala#L51).   
Should we allow a similar abstraction or even provide a shared interface to automatically save both states and metrics by just supplying *one* repository-like object to the [`VerificationRunBuilder`](https://github.com/awslabs/deequ/blob/e333f4b507323cdd84ae084cc711eae1bab7a590/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L28)?

### DIFF
--- a/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
@@ -105,7 +105,7 @@ class VerificationRunBuilder(val data: DataFrame) {
     * Can be used to efficiently compute metrics for a large dataset if e.g. a new partition is added.
     *
     * @param stateLoader A state loader that loads previously calculated states and
-   *                    allows aggregation with the ones calculated in this run.
+    *                    allows aggregation with the ones calculated in this run.
     */
   def aggregateWith(stateLoader: StateLoader): this.type = {
     this.stateLoader = Option(stateLoader)
--- a/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
@@ -105,7 +105,7 @@ class VerificationRunBuilder(val data: DataFrame) {
     * Can be used to efficiently compute metrics for a large dataset if e.g. a new partition is added.
     *
     * @param stateLoader A state loader that loads previously calculated states and
-   *                    aggregates them with the ones calculated in this run.
+   *                    allows aggregation with the ones calculated in this run.
     */
   def aggregateWith(stateLoader: StateLoader): this.type = {
     this.stateLoader = Option(stateLoader)
--- a/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
@@ -66,7 +66,6 @@ class VerificationRunBuilder(val data: DataFrame) {
 
     stateLoader = verificationRunBuilder.stateLoader
     statePersister = verificationRunBuilder.statePersister
-
   }
 
   /**
@@ -90,8 +89,9 @@ class VerificationRunBuilder(val data: DataFrame) {
   }
 
   /**
-    * Use to automatically save analyzer states.
-    * Enables later aggregation e.g. when a new partition is added to the dataset.
+    * Save analyzer states.
+    * Enables aggregate computation of metrics later, e.g., when a new partition is
+    * added to the dataset.
     *
     * @param statePersister A state persister that saves the computed states for later aggregation
     */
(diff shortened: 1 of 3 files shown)


src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
```diff
@@ -60,6 +63,10 @@ class VerificationRunBuilder(val data: DataFrame) {
     overwriteOutputFiles = verificationRunBuilder.overwriteOutputFiles
     saveCheckResultsJsonPath = verificationRunBuilder.saveCheckResultsJsonPath
     saveSuccessMetricsJsonPath = verificationRunBuilder.saveSuccessMetricsJsonPath
+
+    stateLoader = verificationRunBuilder.stateLoader
+    statePersister = verificationRunBuilder.statePersister
+
```
sscdotopen: remove empty line

src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
```diff
@@ -82,6 +89,29 @@ class VerificationRunBuilder(val data: DataFrame) {
     this
   }
 
+  /**
+    * Use to automatically save analyzer states.
+    * Enables later aggregation e.g. when a new partition is added to the dataset.
```
sscdotopen: Enables aggregate computation of metrics later, e.g., when a new partition is added to the dataset.

src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala
```diff
@@ -82,6 +89,29 @@ class VerificationRunBuilder(val data: DataFrame) {
     this
   }
 
+  /**
+    * Use to automatically save analyzer states.
```
sscdotopen: Save analyzer states.

sscdotopen:
This looks good, my only comments related to the wording. Please also fix the style violation that caused the build to fail.

PR #165, opened 2019-10-30
Add numerical & categorical distance comparison

pengzai6666:
*Issue #, if available:*
The unit tests for KLL Sketches, which tests the probability error, are yet to be implemented.

*Description of changes:*
Add distance comparison functions for numerical & categorical profiles.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Distance.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Distance.scala
@@ -23,10 +23,10 @@ case class CategoricalHistogram(buckets: List[CategoricalHistogramBucket])
 object Distance {
 
     /** Calculate distance of numerical profiles based on KLL Sketches and L-Infinity Distance */
-    def calculateNumericalDistance(
+    def numericalDistance(
       sample1: QuantileNonSample[Double],
       sample2: QuantileNonSample[Double],
-      enoughSample: Boolean = false)
+      correctForLowNumberOfSamples: Boolean = false)
     : Double = {
       val rankMap1 = sample1.getRankMap()
       val rankMap2 = sample2.getRankMap()
@@ -41,14 +41,14 @@ object Distance {
         val cdfDiff = Math.abs(cdf1 - cdf2)
         linfSimple = Math.max(linfSimple, cdfDiff)
       }
-      selectMetrics(linfSimple, n, m, enoughSample)
+      selectMetrics(linfSimple, n, m, correctForLowNumberOfSamples)
     }
 
     /** Calculate distance of categorical profiles based on L-Infinity Distance */
-    def calculateCategoricalDistance(
+    def categoricalDistance(
       sample1: CategoricalHistogram,
       sample2: CategoricalHistogram,
-      enoughSample: Boolean = false)
+      correctForLowNumberOfSamples: Boolean = false)
     : Double = {
 
       var countMap1 = scala.collection.mutable.Map[String, Long]()
@@ -72,18 +72,18 @@ object Distance {
         val cdfDiff = Math.abs(cdf1 - cdf2)
         linfSimple = Math.max(linfSimple, cdfDiff)
       }
-      selectMetrics(linfSimple, n, m, enoughSample)
+      selectMetrics(linfSimple, n, m, correctForLowNumberOfSamples)
     }
 
   /** Select which metrics to compute (linf_simple or linf_robust)
    *  based on whether samples are enough */
-   private def selectMetrics(
+   private[this] def selectMetrics(
      linfSimple: Double,
      n: Double,
      m: Double,
-     enoughSamples: Boolean = false)
+     correctForLowNumberOfSamples: Boolean = false)
    : Double = {
-     if (enoughSamples) {
+     if (correctForLowNumberOfSamples) {
        linfSimple
      } else {
        // This formula is based on  “Two-sample Kolmogorov–Smirnov test"
--- a/src/main/scala/com/amazon/deequ/analyzers/Distance.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Distance.scala
@@ -22,11 +22,11 @@ case class CategoricalHistogram(buckets: List[CategoricalHistogramBucket])
 
 object Distance {
 
-    /** Calculate distance of numerical profiles based on KLL Sketch */
+    /** Calculate distance of numerical profiles based on KLL Sketches and L-Infinity Distance */
     def calculateNumericalDistance(
       sample1: QuantileNonSample[Double],
       sample2: QuantileNonSample[Double],
-      calculateLinfSimple: Boolean = false)
+      enoughSample: Boolean = false)
     : Double = {
       val rankMap1 = sample1.getRankMap()
       val rankMap2 = sample2.getRankMap()
@@ -41,21 +41,14 @@ object Distance {
         val cdfDiff = Math.abs(cdf1 - cdf2)
         linfSimple = Math.max(linfSimple, cdfDiff)
       }
-      if (calculateLinfSimple) {
-        linfSimple
-      } else {
-        // This formula is based on  “Two-sample Kolmogorov–Smirnov test"
-        // Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test
-        val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m)))
-        linfRobust
-      }
+      selectMetrics(linfSimple, n, m, enoughSample)
     }
 
     /** Calculate distance of categorical profiles based on L-Infinity Distance */
-    def calculateLInfinityCategoricalDistance(
+    def calculateCategoricalDistance(
       sample1: CategoricalHistogram,
       sample2: CategoricalHistogram,
-      calculateLinfSimple: Boolean = false)
+      enoughSample: Boolean = false)
     : Double = {
 
       var countMap1 = scala.collection.mutable.Map[String, Long]()
@@ -79,14 +72,25 @@ object Distance {
         val cdfDiff = Math.abs(cdf1 - cdf2)
         linfSimple = Math.max(linfSimple, cdfDiff)
       }
-      if (calculateLinfSimple) {
-        linfSimple
-      } else {
-        // This formula is based on  “Two-sample Kolmogorov–Smirnov test"
-        // Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test
-        val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m)))
-        linfRobust
-      }
+      selectMetrics(linfSimple, n, m, enoughSample)
     }
+
+  /** Select which metrics to compute (linf_simple or linf_robust)
+   *  based on whether samples are enough */
+   private def selectMetrics(
+     linfSimple: Double,
+     n: Double,
+     m: Double,
+     enoughSamples: Boolean = false)
+   : Double = {
+     if (enoughSamples) {
+       linfSimple
+     } else {
+       // This formula is based on  “Two-sample Kolmogorov–Smirnov test"
+       // Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test
+       val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m)))
+       linfRobust
+     }
+   }
 }
 
--- a/src/main/scala/com/amazon/deequ/analyzers/Distance.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Distance.scala
@@ -16,19 +16,17 @@
 
 package com.amazon.deequ.analyzers
 
-case class CategoricalBucket(value: String, count: Long)
+case class CategoricalHistogramBucket(value: String, count: Long)
 
-case class Categorical(buckets: List[CategoricalBucket])
+case class CategoricalHistogram(buckets: List[CategoricalHistogramBucket])
 
-class Distance {
+object Distance {
 
-    /** Calculate distance of numerical profiles based on KLL Sketch
-     *  This formula is based on  “Two-sample Kolmogorov–Smirnov test"
-     *  Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test */
+    /** Calculate distance of numerical profiles based on KLL Sketch */
     def calculateNumericalDistance(
       sample1: QuantileNonSample[Double],
       sample2: QuantileNonSample[Double],
-      isSimple: Boolean = false)
+      calculateLinfSimple: Boolean = false)
     : Double = {
       val rankMap1 = sample1.getRankMap()
       val rankMap2 = sample2.getRankMap()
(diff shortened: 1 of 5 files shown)


src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      val rankMap1 = sample1.getRankMap()
+      val rankMap2 = sample2.getRankMap()
+      val combinedKeys = rankMap1.keySet.union(rankMap2.keySet)
+      val n = rankMap1.valuesIterator.max.toDouble
+      val m = rankMap2.valuesIterator.max.toDouble
+      var linfSimple = 0.0
+
+      combinedKeys.foreach { key =>
+        linfSimple = Math.max(linfSimple,
+          Math.abs(sample1.getRank(key, rankMap1) / n - sample2.getRank(key, rankMap2) / m))
```
sscdotopen: I would move this out of the max expression to make it more readable

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+      val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m) ))
```
sscdotopen: I would move this into the else branch to only compute it if necessary (not sure if the compiler is smart enough to automatically do this)

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+      buckets2.foreach { bucket =>
+        m += bucket.count
+        countMap2 += (bucket.value -> bucket.count)
+      }
+      val combinedKeys = countMap1.keySet.union(countMap2.keySet)
+      var linfSimple = 0.0
+      combinedKeys.foreach { key =>
+        linfSimple = Math.max(linfSimple,
+          Math.abs(countMap1.getOrElse(key, 0L) / n - countMap2.getOrElse(key, 0L) / m))
```
sscdotopen: Again, move this out of the max expression to make the code more readable

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+      val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m) ))
```
sscdotopen: again compute this only in the else branch

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      if (isSimple) {
+        linfSimple
+      } else {
+        linfRobust
+      }
+    }
+}
+
+object Distance{
```
sscdotopen: Why do we need this? We can just make the upper code an object, as the Distance class has no state anyways

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
+ */
+
+package com.amazon.deequ.analyzers
+
+case class CategoricalBucket(value: String, count: Long)
+
+case class Categorical(buckets: List[CategoricalBucket])
```
sscdotopen: I find the name Categorical a bit confusing, is this not simply a Histogram?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m) ))
+      if (isSimple) {
+        linfSimple
+      } else {
+        linfRobust
+      }
+    }
+
+    /** Calculate distance of categorical profiles based on L-Infinity Distance */
+    def calculateCategoricalDistance(
```
sscdotopen: could we rename this to reflect that it computes the l-infty distance?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      sample1: Categorical,
+      sample2: Categorical,
+      isSimple: Boolean = false)
+    : Double = {
+
+      val buckets1 = sample1.buckets
+      val buckets2 = sample2.buckets
+      var countMap1 = scala.collection.mutable.Map[String, Long]()
```
sscdotopen: why do we need these copies? can't we simply iterate over the original buckets?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
sscdotopen: Why do we use a list of key-value pairs and not a hashmap? It seems this is used like a hashmap later on

src/main/scala/com/amazon/deequ/analyzers/QuantileNonSample.scala
```diff
@@ -42,6 +42,22 @@ class QuantileNonSample[T](
 
   expand()
 
+  /** Given sketchSize, shrinkingFactor, and data, reconstruct the KLL Object */
+  def reconstruct(k: Int, c: Double, data: Array[Array[T]]): Unit = {
```
sscdotopen: why do we use different names for sketchSize (k), and shrinkingFactor (c)?

src/main/scala/com/amazon/deequ/analyzers/QuantileNonSample.scala
```diff
@@ -42,6 +42,22 @@ class QuantileNonSample[T](
...
+    sketchSize = k
+    shrinkingFactor = c
+    compactors = ArrayBuffer[NonSampleCompactor[T]]()
```
sscdotopen: can we already give this a hint for the right capacity? it seems it needs to hold max(data.indices) values

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+
+class Distance {
+
+    /** Calculate distance of numerical profiles based on KLL Sketch
+     *  This formula is based on  “Two-sample Kolmogorov–Smirnov test"
+     *  Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test */
+    def calculateNumericalDistance(
+      sample1: QuantileNonSample[Double],
+      sample2: QuantileNonSample[Double],
+      isSimple: Boolean = false)
```
sscdotopen: can we give this a more speaking name? It is unclear to me what a simple distance would be

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+      } else {
+        linfRobust
+      }
+    }
+
+    /** Calculate distance of categorical profiles based on L-Infinity Distance */
+    def calculateCategoricalDistance(
+      sample1: Categorical,
+      sample2: Categorical,
+      isSimple: Boolean = false)
```
sscdotopen: can we give this a more speaking name? It is unclear to me what a simple distance would be

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
sscdotopen: Can you add a comment on what is happening here?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
sscdotopen: Can you add a comment on what is happening here?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,90 @@
...
+
+case class CategoricalBucket(value: String, count: Long)
+
+case class Categorical(buckets: List[CategoricalBucket])
+
+class Distance {
+
+    /** Calculate distance of numerical profiles based on KLL Sketch
+     *  This formula is based on  “Two-sample Kolmogorov–Smirnov test"
+     *  Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test */
```
sscdotopen: Can you explain how this is related to a KS test?

sscdotopen:
Looking forward to having this as part of Deequ!

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Sure! Very good suggestion!

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Sure, good point!

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Sure.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Sure.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Yeah this is absolutely correct. I should have thought about that. Fixed.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Yeah this is simply a histogram, it is named "Categorical" here because this is named in our codebases before I work on this project, so I choose to stick to its name. But I guess I could refactor this in Deequ, it's not a big deal. Have Renamed to "CategoricalHistogram".

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Since distance of KLL Sketch is also related to the L-Infinity distance, I would rename it to "calculateLInfinityCategoricalDistance"

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Yeah you are definitely right on this. Have removed these copies.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Again, I would actually like to design it as a hashmap if I could, but for historic reasons in our codebases, and because we are going to use these data structures as input later, I would be inclined to keep it as it is.

src/main/scala/com/amazon/deequ/analyzers/QuantileNonSample.scala
pengzai6666: You are right, it could cause confusions. Fixed.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: This is to decide whether we would calculate the distance for simple L-infinity distance, or robust L-infinity distance. If we would like to calculate simple L-infinity distance, we wouldn't take the number of samples (which is n and m) into calculation (a.k.a simple mode). Renamed it to "calculateLinfSimple".

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Same as above. Resolved.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Added. This comment is originally at the beginning of the function, but now I will put it just before the formula of linf_robust.
"This formula is based on  “Two-sample Kolmogorov–Smirnov test"
 Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test"

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Same as above.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Absolutely. The formula of linf_robust is based on the “Two-sample Kolmogorov–Smirnov test", which is used to test whether two underlying one-dimensional probability distributions differ. Check out the “Two-sample Kolmogorov–Smirnov test" section of the reference link and you could find out where the item 1.8 * Math.sqrt((n + m) / (n * m)) in the formula is coming from (when alpha = 0.001, c(alpha) is approximate to 1.8).

pengzai6666:
Hi Sebastian,
Thank you very much for giving valuable advice with regard to my code! 
I have invited Zohar, who works as a scientist and suggested me on how to do the actual implementation of L-Infinity distance, to take a look at this pull request and my implementation. Moreover, I added a new unit test for KLL Sketches, please feel free to take a look & make comments.

src/main/scala/com/amazon/deequ/analyzers/QuantileNonSample.scala
```diff
@@ -42,6 +42,23 @@ class QuantileNonSample[T](
 
   expand()
 
+  /** Given sketchSize, shrinkingFactor, and data, reconstruct the KLL Object */
+  def reconstruct(
+    sketchSizeInput: Int,
```
sscdotopen: you can just name these sketchSize etc and later use `this.sketchSize =  sketchSize` for disambiguation

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,92 @@
...
+      val m = rankMap2.valuesIterator.max.toDouble
+      var linfSimple = 0.0
+
+      combinedKeys.foreach { key =>
+        val cdf1 = sample1.getRank(key, rankMap1) / n
+        val cdf2 = sample2.getRank(key, rankMap2) / m
+        val cdfDiff = Math.abs(cdf1 - cdf2)
+        linfSimple = Math.max(linfSimple, cdfDiff)
+      }
+      if (calculateLinfSimple) {
```
sscdotopen: the whole if else block is identical between the two methods and can be refactored into its own method

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,92 @@
...
+package com.amazon.deequ.analyzers
+
+case class CategoricalHistogramBucket(value: String, count: Long)
+
+case class CategoricalHistogram(buckets: List[CategoricalHistogramBucket])
+
+object Distance {
+
+    /** Calculate distance of numerical profiles based on KLL Sketch */
+    def calculateNumericalDistance(
```
sscdotopen: I am a still bit confused by this method, does it also compute the L-infty distance?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,92 @@
...
+        val cdf2 = sample2.getRank(key, rankMap2) / m
+        val cdfDiff = Math.abs(cdf1 - cdf2)
+        linfSimple = Math.max(linfSimple, cdfDiff)
+      }
+      if (calculateLinfSimple) {
+        linfSimple
+      } else {
+        // This formula is based on  “Two-sample Kolmogorov–Smirnov test"
+        // Reference: https://en.m.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test
+        val linfRobust = Math.max(0.0, linfSimple - 1.8 * Math.sqrt((n + m) / (n * m)))
```
sscdotopen: If I understand things correctly, then the "robust" computation somehow conducts a hypothesis test, right? It would be great if you could elaborate a bit more on that. I think we should also rename the `calculateLinfSimple` parameter to reflect this.

pengzai6666: Yes you are absolutely right, the "robust" formula conducts a hypothesis test and the formula is based on null hypothesis.  If we have enough samples from two distributions, we could simply calculate ${linfSimple}, but when there aren't enough samples, ${linfSimple} could be large not because the distributions are different but because we don't have enough observations. That's why we make it statistically robust by subtracting 1.8 * Math.sqrt((n + m) / (n * m)), where n and m are numbers of samples in these two distributions. Therefore I suggest to change the variable name from "calculateLinfSimple" to "enoughSamples" to reflect this.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,96 @@
...
+
+case class CategoricalHistogram(buckets: List[CategoricalHistogramBucket])
+
+object Distance {
+
+    /** Calculate distance of numerical profiles based on KLL Sketches and L-Infinity Distance */
+    def calculateNumericalDistance(
+      sample1: QuantileNonSample[Double],
+      sample2: QuantileNonSample[Double],
+      enoughSample: Boolean = false)
```
sscdotopen: can we maybe call this `correctForLowNumberOfSamples` ?

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,96 @@
...
+        val cdf1 = sample1.getRank(key, rankMap1) / n
+        val cdf2 = sample2.getRank(key, rankMap2) / m
+        val cdfDiff = Math.abs(cdf1 - cdf2)
+        linfSimple = Math.max(linfSimple, cdfDiff)
+      }
+      selectMetrics(linfSimple, n, m, enoughSample)
+    }
+
+    /** Calculate distance of categorical profiles based on L-Infinity Distance */
+    def calculateCategoricalDistance(
```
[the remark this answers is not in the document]
pengzai6666: Done. Thanks.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,96 @@
...
+        val cdf2 = countMap2.getOrElse(key, 0L) / m
+        val cdfDiff = Math.abs(cdf1 - cdf2)
+        linfSimple = Math.max(linfSimple, cdfDiff)
+      }
+      selectMetrics(linfSimple, n, m, enoughSample)
+    }
+
+  /** Select which metrics to compute (linf_simple or linf_robust)
+   *  based on whether samples are enough */
+   private def selectMetrics(
```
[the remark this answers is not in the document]
pengzai6666: Done.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
pengzai6666: Sure.

src/main/scala/com/amazon/deequ/analyzers/Distance.scala
```diff
@@ -0,0 +1,96 @@
...
+        linfSimple = Math.max(linfSimple, cdfDiff)
+      }
+      selectMetrics(linfSimple, n, m, enoughSample)
+    }
+
+    /** Calculate distance of categorical profiles based on L-Infinity Distance */
+    def calculateCategoricalDistance(
+      sample1: CategoricalHistogram,
+      sample2: CategoricalHistogram,
+      enoughSample: Boolean = false)
```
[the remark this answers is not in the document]
pengzai6666: Sure.

pengzai6666:
Hi Sebastian,
Please take a last look and make sure that you are fine with the changes. Thanks!

sscdotopen:
Merged, very excited to see this contribution!

PR #166, opened 2019-10-31
Change categorical data structure to map

pengzai6666:
*Issue #, if available:*

*Description of changes:*
Previously we use Categorial case class for distance computation, now change that to mutable map.


pengzai6666:
Hi Sebastian, 
This is a minor code change, which uses mutable map as input to calculate categorical L-Infinity Distance. This change is also suggested by you in the comments from the last pull request. Hope you could take a look & make comments :)

PR #176, opened 2019-12-11
Set predefined data type to respect customer's customized data type in baseline

pengzai6666:
*Issue #, if available:*

*Description of changes:*
This pull request is mainly focusing on respecting the data types customers set in baseline constraints. One simple example would be that there's one column named "Zip Code", customers would like it to have dataType "String", but if the dataType in df.schema is StringType, then Deequ would infer the dataType to be "Integral" and run a numerical analysis. After this change, if a baseline dataType is given for a column, it will be respected. I am making the parameter to be optional so that it won't affect the cases when a baseline is not given.

### DIFF
--- a/src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
+++ b/src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
@@ -119,7 +119,10 @@ object ColumnProfiler {
 
     // We compute completeness, approximate number of distinct values
     // and type detection for string columns in the first pass
-    val analyzersForGenericStats = getAnalyzersForGenericStats(data.schema, relevantColumns)
+    val analyzersForGenericStats = getAnalyzersForGenericStats(
+      data.schema,
+      relevantColumns,
+      predefinedTypes)
 
     var analysisRunnerFirstPass = AnalysisRunner
       .onData(data)
@@ -216,7 +219,8 @@ object ColumnProfiler {
 
   private[this] def getAnalyzersForGenericStats(
       schema: StructType,
-      relevantColumns: Seq[String])
+      relevantColumns: Seq[String],
+      predefinedTypes: Map[String, DataTypeInstances.Value])
     : Seq[Analyzer[_, Metric[_]]] = {
 
     schema.fields
@@ -225,7 +229,7 @@ object ColumnProfiler {
 
         val name = field.name
 
-        if (field.dataType == StringType) {
+        if (field.dataType == StringType && !predefinedTypes.contains(name)) {
           Seq(Completeness(name), ApproxCountDistinct(name), DataType(name))
         } else {
           Seq(Completeness(name), ApproxCountDistinct(name))
@@ -366,7 +370,8 @@ object ColumnProfiler {
     val inferredTypes = results.metricMap
       .filterNot{
         case (analyzer: DataType, _) => predefinedTypes.contains(analyzer.column)
-        case _ => true}
+        case _ => true
+      }
       .collect { case (analyzer: DataType, metric: HistogramMetric) =>
           val typeHistogram = metric.value.get
           analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
@@ -375,7 +380,8 @@ object ColumnProfiler {
     val typeDetectionHistograms = results.metricMap
       .filterNot{
         case (analyzer: DataType, _) => predefinedTypes.contains(analyzer.column)
-        case _ => true}
+        case _ => true
+      }
       .collect { case (analyzer: DataType, metric: HistogramMetric) =>
           val typeCounts = metric.value.get.values
             .map { case (key, distValue) => key -> distValue.absolute }
--- a/src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
+++ b/src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
@@ -34,20 +34,11 @@ private[deequ] case class GenericColumnStatistics(
     typeDetectionHistograms: Map[String, Map[String, Long]],
     approximateNumDistincts: Map[String, Long],
     completenesses: Map[String, Double],
-    predefinedColumnDataTypes: Option[Map[String, String]]) {
+    predefinedColumnDataTypes: Option[Map[String, DataTypeInstances.Value]]) {
 
   def typeOf(column: String): DataTypeInstances.Value = {
     if (predefinedColumnDataTypes.isDefined && predefinedColumnDataTypes.get.contains(column)) {
-      val predefinedDataType = predefinedColumnDataTypes.get(column) match {
-        case "Integral" => Integral
-        case "Fractional" => Fractional
-        case "String" => String
-        case "Boolean" => Boolean
-        case _ =>
-          println(s"Unable to map predefined dataType ${predefinedColumnDataTypes.get(column)}")
-          Unknown
-      }
-      predefinedDataType
+      predefinedColumnDataTypes.get(column)
     } else {
       val inferredAndKnown = inferredTypes ++ knownTypes
       inferredAndKnown(column)
@@ -375,10 +366,22 @@ object ColumnProfiler {
       .head
       .toLong
 
-    val inferredTypes = results.metricMap
-      .collect { case (analyzer: DataType, metric: HistogramMetric) =>
-        val typeHistogram = metric.value.get
-        analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
+
+    val inferredTypes =
+      if (predefinedColumnDataTypes.isDefined) {
+        results.metricMap
+          .collect { case (analyzer: DataType, metric: HistogramMetric)
+            if (!predefinedColumnDataTypes.get.contains(analyzer.column)) =>
+            // only infer dataType of a column when it is not predefined
+              val typeHistogram = metric.value.get
+              analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
+          }
+      } else {
+        results.metricMap
+          .collect { case (analyzer: DataType, metric: HistogramMetric) =>
+            val typeHistogram = metric.value.get
+            analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
+          }
       }
 
     val typeDetectionHistograms = results.metricMap
@@ -399,26 +402,81 @@ object ColumnProfiler {
         analyzer.column -> metric.value.get
       }
 
-    val knownTypes = schema.fields
-      .filter { column => columns.contains(column.name) }
-      .filter { _.dataType != StringType }
-      .map { field =>
-        val knownType = field.dataType match {
-          case ShortType | LongType | IntegerType => Integral
-          case DecimalType() | FloatType | DoubleType => Fractional
-          case BooleanType => Boolean
-          case TimestampType => String  // TODO We should have support for dates in deequ...
-          case _ =>
-            println(s"Unable to map type ${field.dataType}")
-            Unknown
-        }
+    val knownTypes =
+      if (predefinedColumnDataTypes.isDefined) {
+        schema.fields
+          .filter { column => columns.contains(column.name) }
+          .filter { column => !predefinedColumnDataTypes.get.contains(column.name)}
+          .filter {
+            _.dataType != StringType
+          }
+          .map { field =>
+            val knownType = field.dataType match {
+              case ShortType | LongType | IntegerType => Integral
+              case DecimalType() | FloatType | DoubleType => Fractional
+              case BooleanType => Boolean
+              case TimestampType => String // TODO We should have support for dates in deequ...
+              case _ =>
+                println(s"Unable to map type ${field.dataType}")
+                Unknown
+            }
 
-        field.name -> knownType
+            field.name -> knownType
+          }
+          .toMap
+      } else {
+        schema.fields
+          .filter { column => columns.contains(column.name) }
+          .filter {
+            _.dataType != StringType
+          }
+          .map { field =>
+            val knownType = field.dataType match {
+              case ShortType | LongType | IntegerType => Integral
+              case DecimalType() | FloatType | DoubleType => Fractional
+              case BooleanType => Boolean
+              case TimestampType => String // TODO We should have support for dates in deequ...
+              case _ =>
+                println(s"Unable to map type ${field.dataType}")
+                Unknown
+            }
+
+            field.name -> knownType
+          }
+          .toMap
       }
-      .toMap
 
+    // convert predefinedDataTypes from Option[Map[String, String]]
+    // to Option[Map[String, DataTypeInstances.Value]]
+    val predefinedTypes =
+      if (predefinedColumnDataTypes.isDefined) {
+        val predefinedDataTypeStringMap = predefinedColumnDataTypes.get
+        val predefinedConvertedMap =
+        predefinedDataTypeStringMap.collect {
+          case (column: String, predefinedDataTypeString: String) =>
+          val predefinedDataType = predefinedDataTypeString match {
+            case "Integral" => Integral
+            case "Fractional" => Fractional
+            case "String" => String
+            case "Boolean" => Boolean
+            case "Unknown" => Unknown
+            case _ =>
+              val message = s"Unable to map predefined dataType " +
+                s"${predefinedColumnDataTypes.get(column)} for column ${column}. " +
+                s"Currently Deequ only accepts these baseline dataType: " +
+                s"Integral, Fractional, String, Boolean, Unknown." +
+                s"Please double check your baseline dataType."
+              println(message)
+              throw new Exception(message)
+          }
+          column -> predefinedDataType
+        }
+        Some(predefinedConvertedMap)
+      } else {
+        None
+      }
     GenericColumnStatistics(numRecords, inferredTypes, knownTypes, typeDetectionHistograms,
-      approximateNumDistincts, completenesses, predefinedColumnDataTypes)
+      approximateNumDistincts, completenesses, predefinedTypes)
   }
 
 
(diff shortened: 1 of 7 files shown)


sscdotopen:
I think that there is a misunderstanding here. Deequ infers synactic data types (e.g. interpreting a zip code as numeric). What you are talking about is a semantic data type, e.g., the idea that zip codes represent categorical data.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -33,11 +33,25 @@ private[deequ] case class GenericColumnStatistics(
     knownTypes: Map[String, DataTypeInstances.Value],
     typeDetectionHistograms: Map[String, Map[String, Long]],
     approximateNumDistincts: Map[String, Long],
-    completenesses: Map[String, Double]) {
+    completenesses: Map[String, Double],
+    predefinedColumnDataTypes: Option[Map[String, String]]) {
```
sscdotopen: Why is a map from String to String and not from String to DataTypeInstances.Value?

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -33,11 +33,25 @@ private[deequ] case class GenericColumnStatistics(
...
-    val inferredAndKnown = inferredTypes ++ knownTypes
-    inferredAndKnown(column)
+    if (predefinedColumnDataTypes.isDefined && predefinedColumnDataTypes.get.contains(column)) {
+      val predefinedDataType = predefinedColumnDataTypes.get(column) match {
+        case "Integral" => Integral
+        case "Fractional" => Fractional
+        case "String" => String
+        case "Boolean" => Boolean
+        case _ =>
+          println(s"Unable to map predefined dataType ${predefinedColumnDataTypes.get(column)}")
```
sscdotopen: Throw an exception here

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunner.scala
```diff
@@ -70,13 +70,17 @@ class ConstraintSuggestionRunner {
...
-      kllParameters: Option[KLLParameters])
+      kllWrapper: List[Any])
     : ConstraintSuggestionResult = {
 
     // get testset related data from wrapper
     val testsetRatio: Option[Double] = testsetWrapper(0).asInstanceOf[Option[Double]]
     val testsetSplitRandomSeed: Option[Long] = testsetWrapper(1).asInstanceOf[Option[Long]]
 
+    val kllParameters: Option[KLLParameters] = kllWrapper(0).asInstanceOf[Option[KLLParameters]]
+    val predefinedColumnDataTypes: Option[Map[String, String]] =
```
sscdotopen: What exactly does the kllWrapper and the predefined datatypes have to do with each other?

sscdotopen:
If a user specifies a datatype, then deequ should not try to infer it. That is missing from the PR at the moment.

pengzai6666:
What you said is absolutely correct. The tricky thing is that semantically customers would like the zip code to be string, so we need to run the analysis that customers would like to have. It could also include the cases when customers would like a column to have Integral type in baseline but give a Fractional column, then in this case we should respect this and give an integral analysis (e.g. min and max should be integral in numericColumnProfile), but of course deequ will give a dataType violation when a dataType check is added for that column.  Please feel free to correct me if I am wrong here or if there is a better approach.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: DataTypeInstances is private to Deequ packages so I could only pass in map from String to String in our spark app. But I guess I could convert it to be from String to DataTypeInstances.Value in extractGenericStatistics method.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: Yep, good catch, will do.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
sscdotopen: Might make sense to expose it to the outside (e.g., remove the private scope)

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunner.scala
pengzai6666: The kllWrapper, which is similar to testsetWrapper listed above, are simply workarounds to pass the argcount requirements when building the whole package. Otherwise I could delete these wrappers and pass the parameters themselves.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: I decide to respect this private scope (because I am not sure if there are other reasons why it is set to private) and send a new commit, please feel free to review it.

pengzai6666: If you think that it could be exposed, then I could update the code accordingly.

pengzai6666:
Sorry to interrupt but would u mind reviewing my latest commit? It will help me stay unblocked on my project. That said, please take ur time on this. Thanks!

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -133,7 +139,11 @@ object ColumnProfiler {
 
     val firstPassResults = analysisRunnerFirstPass.run()
 
-    val genericStatistics = extractGenericStatistics(relevantColumns, data.schema, firstPassResults)
+    val genericStatistics = extractGenericStatistics(
```
sscdotopen: no need to change the formatting here

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -347,18 +357,31 @@ object ColumnProfiler {
...
       .collect { case (_: Size, metric: DoubleMetric) => metric.value.get }
       .head
       .toLong
 
-    val inferredTypes = results.metricMap
-      .collect { case (analyzer: DataType, metric: HistogramMetric) =>
-        val typeHistogram = metric.value.get
-        analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
+
+    val inferredTypes =
```
sscdotopen: This can be simplified. Also, we should not even compute the typehistogram for a column if we do not want to infer its type.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -379,26 +402,81 @@ object ColumnProfiler {
...
-        val knownType = field.dataType match {
-          case ShortType | LongType | IntegerType => Integral
-          case DecimalType() | FloatType | DoubleType => Fractional
-          case BooleanType => Boolean
-          case TimestampType => String  // TODO We should have support for dates in deequ...
-          case _ =>
-            println(s"Unable to map type ${field.dataType}")
-            Unknown
-        }
+    val knownTypes =
```
sscdotopen: I think this can also be simplified, as it duplicates a lot of code

src/main/scala/com/amazon/deequ/profiles/ColumnProfilerRunBuilder.scala
```diff
@@ -106,11 +108,27 @@ class ColumnProfilerRunBuilder(val data: DataFrame) {
...
+   * Set KLL parameters.
+   *
+   * @param kllParameters kllParameters(sketchSize, shrinkingFactor, numberOfBuckets)
+   */
   def setKLLParameters(kllParameters: Option[KLLParameters]): this.type = {
     this.kllParameters = kllParameters
     this
   }
 
+  /**
```
sscdotopen: Let us use enums instead of strings for the types

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: The only reason that I am changing the format here is to avoid the "line length exceeding 100" violation when building.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: Good point, it could be simplified, for now inferredType will only compute typeHistogram for cases when predefinedDataTypes is not defined or when the column is not contained in the keySet of predefinedDataTypes. I also do similar check in typeDetectionHistograms as well.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: Sure, will modify to avoid duplicating code.

src/main/scala/com/amazon/deequ/profiles/ColumnProfilerRunBuilder.scala
pengzai6666: How about using the ConstrainableDataTypes which is already defined in Deequ? It is public function so that spark app could use it too.

sscdotopen:
One more thought: I think the code can be simplified if we use 
`predefinedColumnDataTypes: Map[String, DataTypeInstances.Value]` instead of `predefinedColumnDataTypes: Option[Map[String, DataTypeInstances.Value]]`

Just make it empty by default (which denotes that there are no predefined types). This lets you avoid a lot of .isDefined calls.

pengzai6666:
Yep that's a good point to make the map empty by default. BTW,  in order to simplify the code is it possible that we could remove the private modifier for DataTypeInstances? It could simplify the code even more by just a small change.

sscdotopen:
Yes

pengzai6666:
Cool, thanks. Will address that.

pengzai6666:
The code has simplified a lot because of your suggestions. Thanks!

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -347,26 +354,29 @@ object ColumnProfiler {
...
       .head
       .toLong
 
+
     val inferredTypes = results.metricMap
-      .collect { case (analyzer: DataType, metric: HistogramMetric) =>
-        val typeHistogram = metric.value.get
-        analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
+      .collect { case (analyzer: DataType, metric: HistogramMetric)
+        if !predefinedColumnDataTypes.contains(analyzer.column) =>
```
sscdotopen: please use a filter before the collect

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -347,26 +354,29 @@ object ColumnProfiler {
...
       }
 
     val typeDetectionHistograms = results.metricMap
-      .collect { case (analyzer: DataType, metric: HistogramMetric) =>
-        val typeCounts = metric.value.get.values
-          .map { case (key, distValue) => key -> distValue.absolute }
-
-        analyzer.column -> typeCounts
+      .collect { case (analyzer: DataType, metric: HistogramMetric)
+        if !predefinedColumnDataTypes.contains(analyzer.column) =>
```
sscdotopen: please use a filter before the collect

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -381,13 +391,16 @@ object ColumnProfiler {
 
     val knownTypes = schema.fields
       .filter { column => columns.contains(column.name) }
-      .filter { _.dataType != StringType }
+      .filter { column => !predefinedColumnDataTypes.contains(column.name)}
```
sscdotopen: can be simplified via filterNot

src/main/scala/com/amazon/deequ/profiles/ColumnProfilerRunBuilder.scala
```diff
@@ -40,6 +40,8 @@ class ColumnProfilerRunBuilder(val data: DataFrame) {
   protected var saveConstraintSuggestionsJsonPath: Option[String] = None
   protected var saveEvaluationResultsJsonPath: Option[String] = None
   protected var kllParameters: Option[KLLParameters] = None
+  protected var predefinedColumnDataTypes: Map[String, DataTypeInstances.Value] =
```
sscdotopen: please rename this to predefinedTypes (everywhere) and change the default argument to Map.empty

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunBuilder.scala
```diff
@@ -45,6 +45,8 @@ class ConstraintSuggestionRunBuilder(val data: DataFrame) {
   protected var saveConstraintSuggestionsJsonPath: Option[String] = None
   protected var saveEvaluationResultsJsonPath: Option[String] = None
   protected var kllParameters: Option[KLLParameters] = None
+  protected var predefinedColumnDataTypes: Map[String, DataTypeInstances.Value] =
```
sscdotopen: please rename to predefinedTypes and change default argument to Map.empty

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunBuilder.scala
```diff
@@ -220,6 +236,15 @@ class ConstraintSuggestionRunBuilder(val data: DataFrame) {
...
     List[Any](testsetRatio, testsetSplitRandomSeed)
   }
+
+  // implement this wrapper to not violate scalastyle requirement on argcount
+  private def kllWrapper(
+    kllParameters: Option[KLLParameters],
+    predefinedColumnDataTypes: Map[String, DataTypeInstances.Value])
+  : List[Any] = {
+
+    List[Any](kllParameters, predefinedColumnDataTypes)
```
sscdotopen: return ` (Option[KLLParameters], Map[String, DataTypeInstances.Value])` instead of  `List[Any]`

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunner.scala
```diff
@@ -70,13 +70,17 @@ class ConstraintSuggestionRunner {
       cacheInputs: Boolean,
       fileOutputOptions: ConstraintSuggestionFileOutputOptions,
       metricsRepositoryOptions: ConstraintSuggestionMetricsRepositoryOptions,
-      kllParameters: Option[KLLParameters])
+      kllWrapper: List[Any])
```
sscdotopen: Use the tuple from above the remove the casts below

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: Oops! Sorry for not catching this in the first place. Sure will do.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: Will do.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
pengzai6666: Sure.

src/main/scala/com/amazon/deequ/profiles/ColumnProfilerRunBuilder.scala
pengzai6666: Ok!

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunBuilder.scala
pengzai6666: Sure.

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunBuilder.scala
pengzai6666: Okay. Changing testWrapper return type to (Option[Double], Option[Long]) as well.

src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunner.scala
pengzai6666: Sure. Will remove the cast.

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -347,26 +353,33 @@ object ColumnProfiler {
...
-        val typeHistogram = metric.value.get
-        analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
+          val typeHistogram = metric.value.get
+          analyzer.column -> DataTypeHistogram.determineType(typeHistogram)
       }
 
     val typeDetectionHistograms = results.metricMap
+      .filterNot{
+        case (analyzer: DataType, _) => predefinedTypes.contains(analyzer.column)
+        case _ => true}
```
sscdotopen: move the bracket to the next line, please

pengzai6666: Sure thing!

PR #186, opened 2020-01-22
Nested schema support in Predconditions

tdhd:
*Description of changes:*

Prior to this change running checks against nested columns (with `some.column`) would falsely report a missing column and fail the check. With this change, checks can be run against data with nested schemas.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -340,7 +340,9 @@ object Preconditions {
       dataType match {
         // nested type, enqueue all fields in it
         case st: StructType =>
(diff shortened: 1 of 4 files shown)


sscdotopen:
I think we should be careful with nested data. I think that a lot of our Analyzer are not defined for nested data. I would rather throw an error whenever we detect the usage of nested data.

tdhd:
Good point, I tested all of our analyzers and they do not all work out of the box. When projecting to nested columns, spark removes everything but the suffix (`some.structure.column` is selected as `column`).

I will investigate some more to see how we could make this work. In general I think that nested data has a ton of applications and is being used quite frequently, so why not let deequ support that too?

sscdotopen:
The problem is that the semantics of many of our operations, which we translate to aggregation queries in SQL might not be defined for nested data. Maybe it would make more sense to ask users to flatten their data before feeding it to deequ.

Issue #189, opened 2020-02-09
Not able to merge 2 FrequenciesAndNumRows in Histogram analyzer


shashank-sharma:
I am trying to compute Histogram but originally I am interested in Frequency but behind the scene when I pass `saveStatesWith` parameter with `aggregateWith` to Histogram analyzer then it tries to merge its FrequenciesAndNumRows and fails to do it by giving error as:

`org.apache.spark.sql.AnalysisException: cannot resolve '`this.com_amazon_deequ_dq_metrics_count`' given input columns: [this.country, this.count, other.country, other.count]`

Is it possible that this feature is not yet fully implemented ?

sscdotopen:
Hi Shashank, 

This sounds like a bug, as the merge should actually be working. Can you provide a code snippet for us so that we can reproduce the error?

Best,
Sebastian

shashank-sharma:
@sscdotopen 

Error is with: https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala#L364

If I remove `COL_PREFIX` and keep `COUNT_COL = "count"` then it works perfectly fine
Do you think that's the solution ?

Initially I was trying to execute it by:

`Histogram("country").calculate(data, saveStatesWith = Option(stateStore))`
And then using it in:

`Histogram("country").calculate(data, saveStatesWith = Option(stateStore), aggregateWith= Option(secondStateStore))`

sscdotopen:
Found the error and issued a PR for fixing it: https://github.com/awslabs/deequ/pull/190

Thank you for uncovering this!

sscdotopen:
Shoudl work now!

PR #190, opened 2020-02-09
Adjusted name for count column in Histogram

sscdotopen:
This fixes a wrong column name in Histogram as outlined in https://github.com/awslabs/deequ/issues/189


src/test/scala/com/amazon/deequ/analyzers/StateAggregationIntegrationTest.scala
```diff
@@ -246,6 +247,22 @@ class StateAggregationIntegrationTest extends WordSpec with Matchers with SparkC
       assert(resultsFromStates == resultsDirect)
     }
 
+    "not throw errors for the example from https://github.com/awslabs/deequ/issues/189" in withSparkSession { session =>
```
iamsteps: Build seems to fail because of style issue with this line (line too long) - https://travis-ci.org/awslabs/deequ/builds/648070226#L2724

Please fix before shipping

sscdotopen: thx, pushed a fix that hopefully works

Issue #22, opened 2018-09-24
More unit tests for column profiling


sscdotopen:
We should have more tests for the column profiling functionality.

klangner:
Is this still relevant? Are there any specific tests which I can start working on?

sscdotopen:
Dear Krzysztof,

We would be happy to have more tests for the column profiling functionality. We have initial tests here: https://github.com/awslabs/deequ/blob/master/src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala

It would be great if we had some tests on small (but real) public data.

Best,
Sebastian

klangner:
Hi Sebastian,

That's sounds interesting. I can look for a dataset like Titanic and create test cases based on them.
I guess I should add them to the repo, so the tests can be run without any internet access?
Can I added them to the `/test-data` folder? Or any other propositions?
Is uncompressed few MB csv file ok as a test data or it is too big?
I will of course only add datasets if it adds any value to the tests.

Thanks
Krzysztof

sscdotopen:
Yes, we should put the data into our repository to not require internet access. We need to make sure that the license of the data allows that though. Adding an uncompressed file with a couple of MB is totally fine.

lange-labs:
Pull request #191 was merged, resolving.

PR #191, opened 2020-02-11
Unit test based on titanic dataset for column profiling.

klangner:
*Issue #22

*Description of changes:*

New unit test for column profiling based on external file with titanic dataset.
It doesn't fix any problems but I think it can be useful as simple example how to create this kind of tests based on the external datasets. It shows where to put the dataset, how to read it and how to add information about its source. Also it has 891 rows so is bigger than other smaller test cases.

The only not exact value is number of unique values in 'PassengerId` column. should be 891 but is approximated as 887. So probably it is ok.

### DIFF
--- a/src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
+++ b/src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
@@ -402,55 +402,53 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
 
   "return correct profile for the Titanic dataset" in withSparkSession { session =>
     val data = session.read.format("csv")
-                .option("inferSchema", "true")
-                .option("header", "true")
-                .load("test-data/titanic.csv")
-    
+      .option("inferSchema", "true")
+      .option("header", "true")
+      .load("test-data/titanic.csv")
+
     val columnProfiles = ColumnProfiler.profile(data)
 
-    val idProfile = columnProfiles.profiles("PassengerId")
-    assert(idProfile.column == "PassengerId")
-    assert(idProfile.dataType == DataTypeInstances.Integral)
-    assert(idProfile.completeness == 1.0)
-    assert(idProfile.approximateNumDistinctValues == 887) //exact is 891
-    
-    val survivedProfile = columnProfiles.profiles("Survived")
-    assert(survivedProfile.column == "Survived")
-    assert(survivedProfile.dataType == DataTypeInstances.Integral)
-    assert(survivedProfile.completeness == 1.0)
-    assert(survivedProfile.approximateNumDistinctValues == 2)
-    
-    val pclassProfile = columnProfiles.profiles("Pclass")
-    assert(pclassProfile.column == "Pclass")
-    assert(pclassProfile.dataType == DataTypeInstances.Integral)
-    assert(pclassProfile.completeness == 1.0)
-    assert(pclassProfile.approximateNumDistinctValues == 3)
+    val expectedProfiles = List(
+      StandardColumnProfile(
+        "PassengerId",
+        1.0,
+        891,
+        DataTypeInstances.Integral,
+        false,
+        Map(),
+        None),
+      StandardColumnProfile("Survived", 1.0, 2, DataTypeInstances.Integral, false, Map(), None),
+      StandardColumnProfile("Pclass", 1.0, 3, DataTypeInstances.Integral, false, Map(), None),
+      StandardColumnProfile("Name", 1.0, 0, DataTypeInstances.String, true, Map(), None),
+      StandardColumnProfile("Sex", 1.0, 2, DataTypeInstances.String, true, Map(), None),
+      StandardColumnProfile("Ticket", 1.0, 681, DataTypeInstances.String, true, Map(), None),
+      StandardColumnProfile("Fare", 1.0, 0, DataTypeInstances.Fractional, false, Map(), None),
+      StandardColumnProfile("Cabin", 0.22, 0, DataTypeInstances.String, true, Map(), None)
+    )
 
-    val nameProfile = columnProfiles.profiles("Name")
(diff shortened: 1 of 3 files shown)


src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
```diff
@@ -400,4 +400,57 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
...
+                .option("header", "true")
+                .load("test-data/titanic.csv")
+    
+    val columnProfiles = ColumnProfiler.profile(data)
+
+    val idProfile = columnProfiles.profiles("PassengerId")
+    assert(idProfile.column == "PassengerId")
+    assert(idProfile.dataType == DataTypeInstances.Integral)
+    assert(idProfile.completeness == 1.0)
+    assert(idProfile.approximateNumDistinctValues == 887) //exact is 891
```
sscdotopen: This value might vary from run to run, better to check whether its in a given interval

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
```diff
@@ -400,4 +400,57 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
...
+    assert(sexProfile.column == "Sex")
+    assert(sexProfile.dataType == DataTypeInstances.String)
+    assert(sexProfile.completeness == 1.0)
+    assert(sexProfile.approximateNumDistinctValues == 2)
+
+    val ticketProfile = columnProfiles.profiles("Ticket")
+    assert(ticketProfile.column == "Ticket")
+    assert(ticketProfile.dataType == DataTypeInstances.String)
+    assert(ticketProfile.completeness == 1.0)
+    assert(ticketProfile.approximateNumDistinctValues == 710)
```
sscdotopen: again this value might vary

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
```diff
@@ -400,4 +400,57 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
...
 
+  "return correct profile for the Titanic dataset" in withSparkSession { session =>
+    val data = session.read.format("csv")
+                .option("inferSchema", "true")
+                .option("header", "true")
+                .load("test-data/titanic.csv")
+    
+    val columnProfiles = ColumnProfiler.profile(data)
+
+    val idProfile = columnProfiles.profiles("PassengerId")
```
sscdotopen: you could refactor the code to check a column profile into its own method as its repeated many times

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
```diff
@@ -400,4 +400,57 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
     }
   }
 
+  "return correct profile for the Titanic dataset" in withSparkSession { session =>
+    val data = session.read.format("csv")
```
sscdotopen: please use 2 spaces indent here

klangner: ok

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
klangner: Yes, makes sense.

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
```diff
@@ -400,4 +400,55 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
...
+      expectedProfiles: List[ColumnProfile])
+    : Unit = {
+
+    expectedProfiles.foreach { expected =>
+      val actual = actualProfiles(expected.column)
+      val msg = s"""(Column "${expected.column}"")"""
+      assert(actual.dataType == expected.dataType, msg)
+      assert(actual.completeness >= expected.completeness, msg)
+      assert(actual.isDataTypeInferred == expected.isDataTypeInferred, msg)
+      if(expected.approximateNumDistinctValues > 0) {
```
sscdotopen: space after if please

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
```diff
@@ -400,4 +400,55 @@ class ColumnProfilerTest extends WordSpec with Matchers with SparkContextSpec
...
+    val columnProfiles = ColumnProfiler.profile(data)
+
+    val expectedProfiles = List(
+      StandardColumnProfile(
+        "PassengerId",
+        1.0,
+        891,
+        DataTypeInstances.Integral,
+        false,
+        Map(),
```
sscdotopen: Please use Map.empty instead of Map() here and in the subsequent calls

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
klangner: ok

src/test/scala/com/amazon/deequ/profiles/ColumnProfilerTest.scala
klangner: ok

sscdotopen:
Thank you for the contribution!

klangner:
Thank you for the help :-). 
I guess there is no need at the moment for another dataset?
If not I will look for another easy task, or if you think that any specific dataset could be useful just let me know.

sscdotopen:
More tests on more datasets are always welcome, but I dont have a specific dataset in mind. You can alternatively have a look at the open issues and check if there is something interesting for you.

Issue #148, opened 2019-10-03
Row level completess


rambrus:
Hi Team,

We have a use case where we want to collect the following metrics:
_the proportion of rows which contain at least 1 null/missing value_

Do you have any suggestion to cover this use case using *deequ*? 

I have workaround idea:
- add a new `boolean` column (e.g. `has_null`) to the input DataFrame that is indicating whether we have any null / missing value in the given row
- use `Compliance("has_null == true")` analyzer

Is there a better solution to this? Maybe I could write a dynamic predicate for Compliance that iterates all the columns and checks if any of them null.

Thanks in advance,
Robert

sscdotopen:
Hi Robert,

You could generate a SQL predicate like "(column_a IS NULL) OR (column_b IS NULL) OR (column_c IS NULL)" containing all your columns and create a constraint for it with`Check.satisfies`

rambrus:
Cool, thanks for the hint, will give it a try!

rambrus:
Tested your suggestion, working fine, closing this thread!
Thanks

rambrus:
Yeah, I think it makes sense to add it a more generic way. It might be useful to provide the ability to specify a list of columns to be analyzed, instead of including all columns by default.
What do you think?

I will take a look at the API and come up with a proposal.

sscdotopen:
I think it would make for a good default to test all columns, but the check should also provide an option to restrict the columns which cannot be null.

rambrus:
Makes sense

rambrus:
Feature implemented, see the above PR

PR #195, opened 2020-02-26
Feature/issue 148

rambrus:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/148

*Description of changes:*
Implemented the `row level completeness` feature as discussed in [issue #148 ](https://github.com/awslabs/deequ/issues/148).
Added the `isCombinedComplete` and `hasCombinedCompleteness` checks to Check class.

To be honest, I'm not convinced that this is the best approach.

I think we have these alternatives to be considered:
- Overload `isComplete` / `hasCompleteness` checks the way you did in `hasUniqueness` Check to be able to support multiple columns
- Do not add a new Check, just provide some predefined Column conditions the way I did in [ColumnCondition](https://github.com/awslabs/deequ/compare/master...rambrus:feature/issue-148#diff-3239736df413d462a02bbcdbef1228f2) object. In this case, the API would look like:
  ```
	.addCheck(Check.satisfies(isEachNonNull(cols), "Multi-Column Completeness"))
  ```
  
I think it's pretty readable, maybe it would be easier to cover these not-so-frequent use cases by providing column conditions instead of Checks. 

However, the question is: do we want to define to these column conditions in the `deequ` project or leave it to the users to do it on client side?
(I think it would be useful to provide some common column conditions in `deequ`.)

@sscdotopen  Please let me know your thoughts on the above.
Thank you,
Robert


src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -130,6 +131,36 @@ case class Check(
...
   }
 
+  /**
+    * Creates a constraint that asserts on completion in combined set of columns.
+    *
+    * @param columns Columns to run the assertion on
+    * @param hint A hint to provide additional context why a constraint could have failed
+    * @return
+    */
+  def isCombinedComplete(
```
sscdotopen: rename this to 'areComplete'

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -130,6 +131,36 @@ case class Check(
...
+
+  /**
+    * Creates a constraint that assert on completion in combined set of columns.
+    *
+    * @param columns Columns to run the assertion on
+    * @param assertion Function that receives a double input parameter and returns a boolean
+    * @param hint A hint to provide additional context why a constraint could have failed
+    * @return
+    */
+  def hasCombinedCompleteness(
```
sscdotopen: rename this to 'haveCombinedCompleteness'

src/main/scala/com/amazon/deequ/checks/ColumnCondition.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.amazon.deequ.checks
+
+object ColumnCondition {
+  def isNonNull(col: String): String = {
+    s"$col IS NOT NULL"
+  }
+
+  def isEachNonNull(cols: Seq[String]): String = {
```
sscdotopen: Please dont build the predicate via string concatenation, which is error prone. Please use Spark's functions from org.apache.spark.sql.function for constructing this.

src/main/scala/com/amazon/deequ/checks/ColumnCondition.scala
```diff
@@ -0,0 +1,34 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
+ */
+
+package com.amazon.deequ.checks
+
+object ColumnCondition {
```
sscdotopen: Please make this private, as it is only intented for use inside Deequ

src/main/scala/com/amazon/deequ/checks/Check.scala
rambrus: good idea!

src/main/scala/com/amazon/deequ/checks/Check.scala
rambrus: should we include `Combined` in the name? maybe `haveCompleteness` to follow the same convention we used in `areComplete`?

src/main/scala/com/amazon/deequ/checks/ColumnCondition.scala
rambrus: all right, will do

src/main/scala/com/amazon/deequ/checks/ColumnCondition.scala
rambrus: sure

src/main/scala/com/amazon/deequ/checks/Check.scala
sscdotopen: Yes, good point, lets call it haveCompleteness

rambrus: done

src/main/scala/com/amazon/deequ/checks/ColumnCondition.scala
rambrus: done

src/main/scala/com/amazon/deequ/checks/ColumnCondition.scala
rambrus: done

sscdotopen:
Thank you for the contribution!

PR #202, opened 2020-03-10
Feature/deequ 149 entropy

rambrus:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/149

*Description of changes:*
Added filtering support to `Entropy` analyzer, please note that this PR is built on top of https://github.com/awslabs/deequ/issues/149, changelist will be more clear once that PR is merged.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #149, opened 2019-10-03
Grouped metrics


rambrus:
Hi Team,

We have a use case where we want to collect some metrics for a grouped DataFrame, we want to see the collected metrics (e.g. Completeness, ApproxQuantiles, Histogram) separately for every group.
We also want to store the metrics in MetricsRepository.
Any best practices on this?

*I have the following idea:*
- filter a group of the DataFrame, pass it to *deequ* for analyis
- store the metrics in MetricsRepository, using the group value as a tag 
- repeat the above steps for every group

It might be working, but it's pretty time-consuming. Any idea to improve this approach?

Thanks in advance,
Robert

sscdotopen:
Unfortunately, we don't have a good idea how to do this in general, because the number of groups might be very large if you have a large number of distinct values in the field by which you group.

If you know the number of groups in advance, then you can do some workaround by generating a constraint per group with our where filter like this:

```
Check(CheckLevel.Warning, "")
 .isComplete("column_of_interest").where("column_to_group_on = 'groupA'") 
 .isComplete("column_of_interest").where("column_to_group_on = 'groupB'") 
 .isComplete("column_of_interest").where("column_to_group_on = 'groupC'") 
```

However this requires you to know the groups in advance. An advantage is that this does not require shuffling the data and runs in a single pass.

rambrus:
Yeah, I've seen this option, but found that `where` filter is only supported in some Checks (e.g. `hasSize`, `hasCompleteness`)

As I understand, many Checks do not support this feature, e.g.
[ApproxQuantiles](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/ApproxQuantiles.scala)

In my understanding this example is invalid:
```
Check(CheckLevel.Warning, "")
  .hasApproxQuantile("numViews", 0.5, _ <= 10).where("column_to_group_on = 'groupA'")
  .hasApproxQuantile("numViews", 0.5, _ <= 5).where("column_to_group_on = 'groupB'")
  .hasApproxQuantile("numViews", 0.5, _ <= 1).where("column_to_group_on = 'groupC'")
```

I assume the `Check` needs to be an instance of `CheckWithLastConstraintFilterable` to be able to use `where`. Is that correct?

sscdotopen:
Yes, your understanding is right. I am not sure why ApproxQuantiles does not support this, it should be possible IMHO.

rambrus:
[...]

PR #203, opened 2020-03-10
Feature/deequ 149 mutual information

rambrus:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/149

*Description of changes:*
Filtering support on `MutualInformation` analyzer. This PR is built on top of https://github.com/awslabs/deequ/pull/202. 


rambrus:
@sscdotopen Added filtering support to `MutualInformation` analyzer. Please note that `Check.hasMutualInformation` had not been covered by unit tests, added a test case and took the current behavior as a baseline.

The filtering part looks good, however, something's weird here.
For the below input data the `MutualInformation(Seq("att1", "att2"))` calculates the `0.5623` metric value.
```
+------+------+------+
| item | att1 | att2 |
+------+------+------+
|    1 | a    | c    |
|    2 | a    | c    |
|    3 | a    | c    |
|    4 | b    | d    |
+------+------+------+
```

In my understanding the result should be `1` (or at least close to 1), since:
if we know `att1`, `att2` can be inferred without doubt.

What do you think? Do I understand the expected behavior of `MutualInformation` right or I miss something?

sscdotopen:
I think you are right, this looks like a bug to me.

rambrus:
@sscdotopen thanks for the clarification, makes perfect sense. yeah, I was referring to the normalized mutual information. I leave the test case as it is, ready for review.

Issue #201, opened 2020-03-09
VerificationSuite: verifying datatype string on a struct causes corrupt verification run


esil-esil:
Running constraint **.hasDataType("[fieldName]", ConstrainableDataTypes.String))** on a field of (erroneous) datatype struct, causes all other successful constraints to fail as well, i.e. affecting/corrupting all other valid constraint verifications that usually run succesfully. 

Failure symptom is identical to when (attempting) to run constraints on sub-levels of nested data. Running of most constraints (see exceptions below) on nested data also affects/corrupts all otherwise successful contraints in the same verification run.

I do know that constraints on nested data is not officially supported, but a few constraints actually do work on nested data, e.g. pattern matching constraints like e.g. .containsURL ... guessing (not tested) that other pattern matching constraints also do work on nested data. 

In such a corrupted run, the constraint_status is 'Failure' for all constraints (even though only 1 should fail), and the constraint_message (see example below) is almost identical to all constraints.

**Looks as if some error handling is needed, so that only the one erroneous constraint should fail, without affecting all other successful constraints in the same run.**

constraint_message example: 
org.apache.spark.sql.AnalysisException: cannot resolve 'StatefulDataType([fieldName1])' due to data type mismatch: argument 1 requires string type, however, '`[fieldName1]`' is of struct<[... LISTING THE ENTIRE STRUCT, which is loooooong]> type.;;
'Aggregate [sum(cast((verbType#15192 = completed) as bigint)) AS sum(CAST((verbType = completed) AS BIGINT))#15270L, sum(cast(cast(isnotnull(id#15180) as int) as bigint)) AS sum(CAST((id IS N... (chopped off in constraint_message)

sscdotopen:
Thank you, this is a good catch. Deequ does not support nested data (e.g., struct fields). We should extend our preconditions in all Analyzer to check for this. Would you like to work on a PR for that?

esil-esil:
@sscdotopen 
I would love to if I had the competencies, but I am software tester (not developer), so unfortunately, working on a PR is beyond my programming competencies :-/

Any other way we could ask the community for some help regarding a PR for this? E.g. prefixing the headline of this issue, or adding a specific Label to indicate request for help?

As always, thank you very much for your prompt response!

sscdotopen:
We could address this PR together. Would you want to come up with a test case that shows how to produce this error? I can try to implement the fix then.

esil-esil:
@sscdotopen 
Sure, sounds great!

[...]

PR #214, opened 2020-03-23
Additional preconditions to check for nested columns

sscdotopen:
Adds additional preconditions to check for nested columns #201 

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -287,6 +287,8 @@ object Preconditions {
   private[this] val numericDataTypes =
     Set(ByteType, ShortType, IntegerType, LongType, FloatType, DoubleType, DecimalType)
 
+  private[this] val nestedDataTypes = Set(StructType, MapType, ArrayType)
+
   /* Return the first (potential) exception thrown by a precondition */
   def findFirstFailing(
       schema: StructType,
@@ -318,6 +320,18 @@ object Preconditions {
     if (columns.size != n) {
       throw new NumberOfSpecifiedColumnsException(s"$n columns have to be specified! " +
         s"Currently, columns contains only ${columns.size} column(s): ${columns.mkString(",")}!")
+    }
+  }
+
+  def isNotNested(column: String): StructType => Unit = { schema =>
+    if (schema.fieldNames.contains(column)) {
+      val columnDataType = schema(column).dataType
+      columnDataType match {
+        case _ : StructType | _ : MapType | _ : ArrayType =>
+          throw new WrongColumnTypeException(
+            s"Unsupported nested column type of column $column: $columnDataType!")
+        case _ =>
+      }
     }
   }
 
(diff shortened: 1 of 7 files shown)


sscdotopen:
@esil-esil

esil-esil:
@sscdotopen , cc: @rambrus 
Assuming that I still have to wait for a new release of the library to include #149 and #201 (and several other contributions)? Do you have any idea of ETA of new release? Thank you very much in advance!

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -321,6 +323,18 @@ object Preconditions {
     }
   }
 
+  def isNotNested(column: String): StructType => Unit = { schema =>
+    if (schema.fieldNames.contains(column)) {
```
rambrus: We check if the given column exists, but we ignore if it's missing. 

I think there are 2 options here:
- we assume that the existence of the column had been already checked  before this precondition evaluated (like in case of `isNumeric` and `isString`) -> it's a bit unsafe
- in every column predicate (`isNumeric`, `isString` and `isNotNested`) we implicitly call `hasColumn` precondition and make sure the column exists before we do any further verification

What's your view on this?

sscdotopen: We implemented our analyzers to always issue a `hasColumn` check first, so this problem should be taken care of.

sscdotopen:
Updated the PR to include the DataType analyzer as well.

esil-esil:
@sscdotopen , @rambrus 
I know everybody are probably very busy, but I would highly appreciate a new (pre-)release of the Deequ library including especially #149 and #201 . Do you have any idea of ETA of new release? Is there any chance it could be before Easter break?

I have a deadline approaching, and I would be so sad to have to split my complex dataframe into a very high number of atomic sub-grouped-dataframes to mimic the extended filtering already implemented by #149, but not yet released. 

Thank you very much in advance!

sscdotopen:
I asked my colleagues if they have time to cut a release.

rambrus:
@esil-esil: you can clone the repository, build a snapshot jar from `master` branch and install it to your local maven repository using this command:
`mvn install -DskipTests=true -Dscalastyle.skip=true -Dmaven.javadoc.skip=true -B -V`

You will have the SNAPSHOT jar on your local machine, maybe it helps you to get unblocked and start experimenting.

Please note that when I'm trying to compile the project on my local machine, I get this error:
```
ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /C:/work/workspaces/github/rambrus/deequ/src/test/scala/com/amazon/deequ/KLL/KLLBenchmark.java:[38,8] duplicate class: com.amazon.deequ.KLL.KLLBenchmark
[INFO] 1 error
```

I have only seen this error on local machine, `travis ci` seems to be happy.
Fix: delete the file above and it's successfully compiling.
Maybe it's just a issue on my machine. @sscdotopen have you seen this error yet?

esil-esil:
@iamsteps, @sscdotopen : thank you very much for cutting the pre-release 1.0.3-rc2 👍 
@rambrus : thank you for your input, now using new pre-release 😃

Issue #220, opened 2020-04-02
Any Completeness - ColumnCondition.isAnyNotNull


jameswinegar:
I have a use case where it makes sense to check if any of a set of columns have a value. This is the `OR NOT NULL` scenario instead of `AND NOT NULL` implemented in https://github.com/awslabs/deequ/pull/195.

I could take a stab at this tonight, not a Scala expert though.

jameswinegar:
Fixed in #221

PR #221, opened 2020-04-05
Determine any data completeness over a set of columns

jameswinegar:
*Issue #, if available:*
#220 

*Description of changes:*
Adds isAnyNotNull ColumnCondition which allows checking for if data exist in a set of columns. Example use case would be mutually exclusive data in a denormalized table populated with nulls for the column without data.


sscdotopen:
looks good to me, can be merged

Issue #227, opened 2020-04-17
AnalysisRunner passing same exact Analyzer should count as one single Analyzer


shashank-sharma:
Ideally the case should be, if I specify one Analyzer multiple times, then it should consider it as 1 single analyzer, right now output is as intended it gives 1 metric but there is still computation happening behind the scenes

To replicate I have written minimal code as an example which add same analyzer multiple time just to check how much time it took to run (also in code I verified that it is running it)

```
import com.amazon.deequ.analyzers.runners.AnalysisRunner
import com.amazon.deequ.analyzers.Size

case class Item(
  id: Long,
  productName: String,
  description: String,
  priority: String,
  numViews: Long
)

val rdd = spark.sparkContext.parallelize(Seq(
  Item(1, "Thingy A", "awesome thing.", "high", 0),
  Item(2, "Thingy B", "available at http://thingb.com", null, 0),
  Item(3, null, null, "low", 5),
  Item(4, "Thingy D", "checkout https://thingd.ca", "low", 10),
  Item(5, "Thingy E", null, "high", 12)))

val data = spark.createDataFrame(rdd)

def test(count: Int): Unit = {
  val startNanoTime = System.nanoTime()
  val result = AnalysisRunner.onData(data).addAnalyzers(List.fill(count)(Size())).run()
  println(s"${(System.nanoTime() - startNanoTime) / 1000000000.0}")
}

test(1)
test(100)
test(1000)
test(10000)
```

Right now with 1.0.3-RC1 output is:

```
0.094738479
0.658176572
2.00287351
16.243372952
```

After reading code if we make one small change in:

https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala#L113
to
`val allAnalyzers = analyzers.map { _.asInstanceOf[Analyzer[State[_], Metric[_]]] }.distinct`
Then output is
```
0.119323052
0.102618298
0.104985755
0.107584708
```

This can be useful when working with bigger dataset and making use of VerificationSuite as:

```
  val verificationResult = VerificationSuite()
    .onData(data)
    .addRequiredAnalyzer(Size())
    .addCheck(Check(CheckLevel.Error, "size check").hasSize(_ == 5))
    .run() 
```

Here it should compute for size once but in this run at:

https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/VerificationSuite.scala#L119

[...]

PR #228, opened 2020-04-19
Fix duplicate analyzer for AnalysisRunner

shashank-sharma:
*Issue #, if available:* #227 

*Description of changes:*

For AnalysisRunner, now if user gives duplicates analyzers, then it will throw error with list of analyzers which were causing them

In VerificationSuite, as discussed in issue there is case when user can add "Checks" + "RequiredAnalyzer" because of which it creates analyzers list with duplicates and we can't stop it from user side so we take distinct values of it

In AnalysisTest, there was one test case which will fail because of this change hence removed and added new one

### DIFF
--- a/src/main/scala/com/amazon/deequ/VerificationSuite.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationSuite.scala
@@ -116,7 +116,7 @@ class VerificationSuite {
         VerificationFileOutputOptions())
     : VerificationResult = {
 
-    val analyzers = requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }
+    val analyzers = (requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }).distinct
 
     val analysisResults = AnalysisRunner.doAnalysisRun(
       data,
--- a/src/main/scala/com/amazon/deequ/VerificationSuite.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationSuite.scala
@@ -116,7 +116,7 @@ class VerificationSuite {
         VerificationFileOutputOptions())
     : VerificationResult = {
 
-    val analyzers = (requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }).distinct
+    val analyzers = requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }
 
     val analysisResults = AnalysisRunner.doAnalysisRun(
       data,
--- a/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
@@ -111,9 +111,9 @@ object AnalysisRunner {
     }
 
     val allAnalyzers = analyzers.map { _.asInstanceOf[Analyzer[State[_], Metric[_]]] }
-    val distinctAnalyzer = allAnalyzers.distinct
-    require(distinctAnalyzer.size == allAnalyzers.size,
-      s"Duplicate analyzers found: ${allAnalyzers.diff(distinctAnalyzer).distinct}")
+    val distinctAnalyzers = allAnalyzers.distinct
+    require(distinctAnalyzers.size == allAnalyzers.size,
+      s"Duplicate analyzers found: ${allAnalyzers.diff(distinctAnalyzers).distinct}")
 
     /* We do not want to recalculate calculated metrics in the MetricsRepository */
     val resultsComputedPreviously: AnalyzerContext =
--- a/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
@@ -111,6 +111,9 @@ object AnalysisRunner {
     }
 
     val allAnalyzers = analyzers.map { _.asInstanceOf[Analyzer[State[_], Metric[_]]] }
+    val distinctAnalyzer = allAnalyzers.distinct
+    require(distinctAnalyzer.size == allAnalyzers.size,
+      s"Duplicate analyzers found: ${allAnalyzers.diff(distinctAnalyzer).distinct}")
 
     /* We do not want to recalculate calculated metrics in the MetricsRepository */
     val resultsComputedPreviously: AnalyzerContext =
--- a/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
@@ -339,7 +339,7 @@ object AnalysisRunner {
 
     /* Run non-shareable analyzers separately */
     val otherMetrics = others
-      .map { analyzer => analyzer -> analyzer.calculate(data) }
+      .map { analyzer => analyzer -> analyzer.calculate(data, aggregateWith, saveStatesTo) }
       .toMap[Analyzer[_, Metric[_]], Metric[_]]
 
     sharedResults ++ AnalyzerContext(otherMetrics)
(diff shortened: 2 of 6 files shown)


src/main/scala/com/amazon/deequ/VerificationSuite.scala
```diff
@@ -116,7 +116,7 @@ class VerificationSuite {
         VerificationFileOutputOptions())
     : VerificationResult = {
 
-    val analyzers = requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }
+    val analyzers = (requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }).distinct
```
sscdotopen, line 119: this should not be necessary

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
```diff
@@ -341,5 +341,23 @@ class AnalysisRunnerTests extends WordSpec with Matchers with SparkContextSpec w
...
+      val df = getDfWithNumericValues(sparkSession)
+
+      val analyzers = Size() :: Completeness("item") :: Size() :: Nil
+
+      try {
+        AnalysisRunner.onData(df)
+          .addAnalyzers(analyzers)
+          .useSparkSession(sparkSession)
+          .run()
+        assert(false)
```
sscdotopen, line 356: you can use fail() here and remove the asserts

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
```diff
@@ -341,5 +341,23 @@ class AnalysisRunnerTests extends WordSpec with Matchers with SparkContextSpec w
...
       }
     }
+
+    "should give error for duplicate analyzers" in withSparkSession { sparkSession =>
+
+      val df = getDfWithNumericValues(sparkSession)
+
+      val analyzers = Size() :: Completeness("item") :: Size() :: Nil
+
+      try {
```
rambrus, line 351: I recommend to use ScalaTest's `assertThrows` instead of `try / catch` block, like:
```
assertThrows[IllegalArgumentException] { 
     AnalysisRunner.onData(df)
          .addAnalyzers(analyzers)
          .useSparkSession(sparkSession)
          .run()
}
```

See details here: http://www.scalatest.org/user_guide/using_assertions

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
```diff
@@ -341,5 +341,23 @@ class AnalysisRunnerTests extends WordSpec with Matchers with SparkContextSpec w
         inputStream => assert(inputStream.read() > 0)
       }
     }
+
+    "should give error for duplicate analyzers" in withSparkSession { sparkSession =>
```
rambrus, line 345: I would add one more test case to make sure no exception raised when the same Analyzer added with various filtering options, e.g.
`val analyzers = Size() :: Size(Some("att1 = 0")) :: Size(Some("att2 > 0")) :: Nil`

This is a valid case, no exception should be thrown.

src/main/scala/com/amazon/deequ/VerificationSuite.scala
shashank-sharma: So I added this because it was failing multiple checks and one of the main scenarios are:

If requiredAnalyzers is passed with checks during VerificationSuite then it will throw error because it adds analyzers 2 times (1 from analyzers itself and 2 from checks requiredAnalyzers) 
https://github.com/awslabs/deequ/blob/master/src/test/scala/com/amazon/deequ/VerificationResultTest.scala#L184

That's why I took distinct

And also it fails for:
https://github.com/awslabs/deequ/blob/master/src/test/scala/com/amazon/deequ/VerificationSuiteTest.scala#L251

Where multiple anomalyChecks used for same analyzer where Size analyzers is taken multiple times

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
shashank-sharma: Thanks for the suggestion, I have made changes and used intercept for it, does that work ? With assertThrows there is version conflict, not able to find

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
shashank-sharma: Make sense, Added and it is passing

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
shashank-sharma: Fixed with @rambrus suggestion as mentioned below

src/main/scala/com/amazon/deequ/VerificationSuite.scala
shashank-sharma: What if we add the difference between check analyzer and normal analyzer during verificationRun ... This will solve that use case and give error when user specify duplicate analyzers during this run

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
rambrus: ok, looks good

src/main/scala/com/amazon/deequ/VerificationSuite.scala
sscdotopen: I think we should fix the test to not configure duplicate analyzers and leave the code as is. Let us not mangle with the analyzers configured by the users, lets just notify them in case of duplicates.

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
rambrus: @shashank-sharma 
You can use this pattern here to make the code more readable:
```
    noException shouldBe thrownBy {
      AnalysisRunner.onData(df)
        .addAnalyzers(analyzers)
        .useSparkSession(sparkSession)
        .run()
    }
```

src/main/scala/com/amazon/deequ/VerificationSuite.scala
shashank-sharma: So are you suggesting me to fix all those test cases which were failing ? And remove this line with distinct ?

Also is there any problem using it like:
`val analyzers = requiredAnalyzers ++ checks.flatMap { _.requiredAnalyzers() }.diff(requiredAnalyzers)`

If above code snippet works then it will solve everything else I'll fix the test cases

sscdotopen: Please fix the test cases and do not modify the analyzers specified by the users.

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
shashank-sharma: Yup, it is much better, Thanks I have made changes

src/main/scala/com/amazon/deequ/VerificationSuite.scala
shashank-sharma: @sscdotopen I have made necessary changes to test cases and reverted back this code line 

Now all test cases is passing, I have not added/deleted any analyzers but removed those which were added multiple times and causing duplicates error can you please check and let me know if it looks good

src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
```diff
@@ -111,6 +111,9 @@ object AnalysisRunner {
     }
 
     val allAnalyzers = analyzers.map { _.asInstanceOf[Analyzer[State[_], Metric[_]]] }
+    val distinctAnalyzer = allAnalyzers.distinct
```
sscdotopen, line 114: please called this distinctAnalyzers

shashank-sharma: Ah my bad ... Fixed

sscdotopen:
looks good now!

shashank-sharma:
Thanks for the help !

Issue #232, opened 2020-05-01
Extending MetricsRepository for ElasticSearch


timgent:
I'd like to look at extending `MetricsRepository` to store metrics with elastic search. The idea would be to make it easy to graph metrics over time with Kibana. I found some difficulties doing this though and would appreciate any guidance:

**Issues with private AnalysisResult case class constructor**
* It required implementing the load method, which returns a `MetricsRepositoryMultipleResultsLoader`
* Implementing `MetricsRepositoryMultipleResultsLoader` requires a `get` method returning a Seq[AnalysisResult]. AnalysisResult constructor is private (because the object is private it seems I can't access apply method) so I can't construct instance of this

**Query on persisting of Analyzers**
* I also noticed Analyzers currently need to be stored, and the objects themselves are stored with java serialization to a String (I realise in file system implementation everything is stored as serialized java objects). However the only time this is used is when filtering for metrics you want to use (in Check.isNewestPointNonAnomalous). I wondered if it is necessary to persist the Analyzer objects or if we could instead persist some identifier for them so they can be looked up on that? It could make persisting a bit neater to not have the serialized objects stored. Just for ElasticSearch implementation I would have to either take a similar approach for persisting the Analyzers, or have some other way to serialize them to/from JSON

Thoughts appreciated. I am happy to raise PRs for things if it's helpful

sscdotopen:
Sounds like a good direction, please go ahead and issue a PR!

timgent:
@sscdotopen it would be good to get some steer on what PR you would like to see and what it should include. I understand reasons for rejection of the last PR, especially now that I understand the codebase a bit better. And I realise most analyzers you have can be reconstructed perfectly from their JSON representation, just not those with functions in them (which I think is only 1 or 2).

PR #242, opened 2020-05-11
Repositories to use AnalyzerId instead of Analyzer

timgent:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/232

*Description of changes:*
Updating repositories to store a map of analyzerId to metric, rather than the whole analyzer. Idea is that it makes extending repository more straight forward if you want to store metrics somewhere else, as you have some simpler thing to store, rather than storing Analyzer object itself. Should make it simpler to switch file system repository to store data in some more structured format rather than storing as serialized objects (which for example would be hard to do trend analysis with some external tool).

NOTE: This change will break backwards compatibility for people storing metrics with existing method.

### DIFF
--- a/src/main/scala/com/amazon/deequ/VerificationResult.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationResult.scala
@@ -16,12 +16,13 @@
 
 package com.amazon.deequ
 
-import com.amazon.deequ.analyzers.{Analyzer, AnalyzerId}
+import com.amazon.deequ.analyzers.{Analyzer, AnalyzerName}
 import com.amazon.deequ.analyzers.runners.AnalyzerContext
+import com.amazon.deequ.analyzers.runners.AnalyzerContext.SimpleMetricOutput
 import com.amazon.deequ.checks.{Check, CheckResult, CheckStatus}
 import com.amazon.deequ.metrics.Metric
 import com.amazon.deequ.repository.SimpleResultSerde
-import org.apache.spark.sql.{DataFrame, SparkSession}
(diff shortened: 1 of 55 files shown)


timgent:
Realise the code formatting check failed in CI - happy to fix that, though would be good to get feedback on the changes before I put more time into it

sscdotopen:
Dear Tim,

Do you think there is a way to implement this without breaking backwards compatibility? Maybe via an optional parameter or so? It would be to dangerous to merge this otherwise.

timgent:
@sscdotopen 2 ways I can think of to stop issues of backwards compatibility:

1) Could implement a fallback when reading data from the repository. If it fails to read a Map[AnalyzerId, Metric] it could fallback to reading Map[Analyzer, Metric], then just maps the values to get the AnalyzerId

2) Make the interface that a user could extend use AnalyzerId, but within the library convert from AnalyzerId to Analyzer... which feels a bit wrong.

Thoughts?

Also what do you think in terms of if this change is worth considering?

sscdotopen:
Users can already store metrics in JSON form on disk, so I don't think we need to worry about serialized objects.

timgent:
Ahh, looks like I got the wrong understanding of some bits of the code. Thanks for the clarification. I should be able to make this not break backwards compatibility in that case

timgent:
@sscdotopen the change is now backwards compatibility - I added a test for that too just to make sure

timgent:
@sscdotopen any more thoughts on this one?

sscdotopen:
I am not sure about this PR, because we already have a serialisation mechanism in place

PR #258, opened 2020-07-01
update documentation, example:

aviatesk:
*Issue #, if available:*

*Description of changes:*

I did three small udpates:

1. rename `productName` to `manufacturerName` in [albebraic_states_example](https://github.com/awslabs/deequ/blob/22bcff91d6bd8abe7203f696e0fa9d8d0defef57/src/main/scala/com/amazon/deequ/examples/algebraic_states_example.md) doc
It uses `productName` field, but in [its executable](https://github.com/awslabs/deequ/blob/47b300a230213b20ce4e533cf3165c285912f570/src/main/scala/com/amazon/deequ/examples/entities.scala#L29), there is no field `productName` but `manufactureName` is. And so I made them consistent.

2. fix typo in [VerificationRunBuilder.scala](https://github.com/awslabs/deequ/blob/22bcff91d6bd8abe7203f696e0fa9d8d0defef57/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala)

3. fix [KLLExample.scala](https://github.com/awslabs/deequ/blob/47b300a230213b20ce4e533cf3165c285912f570/src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunBuilder.scala)
Currently `ConstraintSuggestionRunner` can't run KLL and so the example causes `None.get` error.
I know `ConstraintSuggesionRunBuilder` [can accept KLL parameters](https://github.com/awslabs/deequ/blob/47b300a230213b20ce4e533cf3165c285912f570/src/main/scala/com/amazon/deequ/suggestions/ConstraintSuggestionRunBuilder.scala#L156-L164) but there is no code to run KLL using that.
Guessing from tests, I think using `ColumnProfilerRunner` is the right way to use KLL, right ?

### DIFF
--- a/src/main/scala/com/amazon/deequ/examples/KLLExample.scala
+++ b/src/main/scala/com/amazon/deequ/examples/KLLExample.scala
@@ -19,7 +19,7 @@ package com.amazon.deequ.examples
 import com.amazon.deequ.analyzers.KLLParameters
 import com.amazon.deequ.examples.ExampleUtils.{itemsAsDataframe, withSpark}
 import com.amazon.deequ.profiles.NumericColumnProfile
-import com.amazon.deequ.suggestions.{ConstraintSuggestionRunner, Rules}
+import com.amazon.deequ.profiles.ColumnProfilerRunner
 
 private[examples] object KLLExample extends App {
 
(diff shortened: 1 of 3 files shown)


src/main/scala/com/amazon/deequ/examples/KLLExample.scala
```diff
@@ -32,13 +32,13 @@ private[examples] object KLLExample extends App {
...
-      .addConstraintRules(Rules.DEFAULT)
-      .setKLLParameters(KLLParameters(2, 0.64, 2))
+      .withKLLProfiling()
+      .setKLLParameters(Some(KLLParameters(2, 0.64, 2)))
       .run()
 
-    val columnProfiles = suggestionResult.columnProfiles
+    val columnProfiles = profileResult.profiles
 
     println("Observed statistics:")
```
aviatesk, line 43: I think we can substitute those printing lines below with the single line `println(ColumnProfiles.toJson(columnProfiles.map { case (_, profile) => profile }.toSeq))`, but not sure you want that.

PR #259, opened 2020-07-02
suggest `CategoricalRangeRule` for integer columns

aviatesk:
*Issue #, if available:*

*Description of changes:*

Currently `(Fractional)CategoricalRangeRule` is only applied to `String`-valued columns (because of [this filtering](https://github.com/awslabs/deequ/blob/47b300a230213b20ce4e533cf3165c285912f570/src/main/scala/com/amazon/deequ/suggestions/rules/CategoricalRangeRule.scala#L30)), as such deequ doesn't suggest any categorical rule for columns whose values are numeric.

But as far as I can tell, read-world data so often have "categorical" columns whose values are actually numeric (mostly integer), e.g. the famous [tatanic](https://www.kaggle.com/c/titanic/data) data also has `Survived` column whose values are `0` or `1`, while it's obviously supposed to be considered as "categorical" (or rather flag) column.
I believe suggesting `isContainedIn` rules for those "categorical-integer" columns as well as usual string-categorical columns would be really valuable (at least for our usage), so I made this PR.

### DIFF
--- a/src/main/scala/com/amazon/deequ/suggestions/rules/CategoricalRangeRule.scala
+++ b/src/main/scala/com/amazon/deequ/suggestions/rules/CategoricalRangeRule.scala
@@ -27,7 +27,8 @@ import org.apache.commons.lang3.StringEscapeUtils
 case class CategoricalRangeRule() extends ConstraintRule[ColumnProfile] {
 
   override def shouldBeApplied(profile: ColumnProfile, numRecords: Long): Boolean = {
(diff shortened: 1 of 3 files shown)


r-aviatesk:
@TammoR can you tag an new release including this, please ?

/cc @sscdotopen

aviatesk:
@TammoR can you tag an new release including this, please ?

/cc @sscdotopen

Issue #262, opened 2020-07-20
RowLevelSchema().withIntColumn is reporting non integers in validRows


dmiljkovic:
I changed line 124 in RowLevelSchemaValidatorTest to be non integer. Code is not reporting that value in invalid results.
Integer value is reported in validRows. I changer 123 to 1.23 and  got 1 in validRows. Below is the code from RowLevelSchemaValidatorTest. Only difference is that 123 is now 1.23

val data = Seq(
        **"1.23", // original val was "123",**
        "N/A",
        "456",
        "999999",
        "-9",
        "-100000",
        null
      ).toDF("id")

      val schema = RowLevelSchema()
        .withIntColumn("id", isNullable = false, minValue = Some(-10), maxValue = Some(1000))

      val result = RowLevelSchemaValidator.validate(data, schema)

      assert(result.numValidRows == 3)
      val validIds = result.validRows.select("id").collect.map { _.getInt(0) }.toSet
      assert(validIds.size == result.numValidRows)
      assert(validIds.contains(1)) // original code assert(validIds.contains(123))
      assert(validIds.contains(456))
      assert(validIds.contains(-9))

tdhd:
Thanks for raising this issue. We will have a look at this shortly.

PR #263, opened 2020-07-20
RowLevelSchemaValidator: Integer constraints for fractional data

tdhd:
*Issue #, if available:*

#262 

*Description of changes:*

In Spark, when casting a string like "1.23" to int the fractional part will simply be ignored (thanks to `UTF8String` in spark-unsafe). This results in a false schema match for these cases. Casting to Decimal avoids this by doing a precision/scale-aware cast.


twollnik:
Thanks for introducing this change, looks useful! Can you add some clarifying comments, I am having a hard time following the exact reasoning in the execution flow.

tdhd:
I propose, we come up with a deequ roadmap and prioritize this (and other issues) accordingly. Closing for now.

PR #307, opened 2020-10-22
Fixes #301: Datatype REGEX fix for column profiler;

### LINKED ISSUE #301: Column Profiler reads String column with all values empty improperly
1. In case all values of a String column are empty ("") then the profiler suggests an Integer data type.
2. In case all values are empty("") then profiler still suggests completeness of 1.0. This might not always be the case for a user. The user may want to check if data is present or not, meaning "" should mean incomplete for the user. We should have flexibility to provide the inference of empty strings ("") as data present or not present from the columnprofiler() point of view.

academy-codex:
Fixes #301 

Changed the REGEX for Integral and Fractional datatype.
This is because if a column has all values as empty string ("") then it is suggested as a integral column with a numeric column profile while it should not be.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/catalyst/StatefulDataType.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/catalyst/StatefulDataType.scala
@@ -33,8 +33,8 @@ private[sql] class StatefulDataType extends UserDefinedAggregateFunction {
   val BOOLEAN_POS = 3
   val STRING_POS = 4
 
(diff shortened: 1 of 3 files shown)


academy-codex:
@sscdotopen Hey Sebastian, CI/CD is throwing this error

**Failed to execute goal org.apache.rat:apache-rat-plugin:0.12:check (default) on project deequ: Too many files with unapproved license**

How do we go about resolving this ?

sscdotopen:
Thank you, this looks great!

PR #331, opened 2020-12-28
fix RetainTypeRule when there're multiple kinds of numeric values

aviatesk:
*Issue #, if available:*

Currently deequ will suggest constraint of either of `Integral` or `Fractional` for `RetainType` rule.
This is problematic when a column holds both integral and fractional values,  because it means 
suggested checks will fail even on the data from which they are generated.

*Description of changes:*

When a column is not pure `Integral` / `Fractional`, now `RetainTypeRule` will suggest 
`.hasDataType(column, Numeric)` constraint, which accepts both type of numeric values.

### DIFF
--- a/src/main/scala/com/amazon/deequ/suggestions/rules/RetainTypeRule.scala
+++ b/src/main/scala/com/amazon/deequ/suggestions/rules/RetainTypeRule.scala
@@ -37,10 +37,21 @@ case class RetainTypeRule() extends ConstraintRule[ColumnProfile] {
   }
 
   override def candidate(profile: ColumnProfile, numRecords: Long): ConstraintSuggestion = {
-
+    val isTypeEmpty =
+      (typ: DataTypeInstances.Value) => profile.typeCounts.getOrElse(typ.toString, 0) == 0
     val typeToCheck = profile.dataType match {
-      case DataTypeInstances.Fractional => ConstrainableDataTypes.Fractional
-      case DataTypeInstances.Integral => ConstrainableDataTypes.Integral
+      case DataTypeInstances.Fractional =>
+        if (isTypeEmpty(DataTypeInstances.Integral)) {
+          ConstrainableDataTypes.Fractional
+        } else {
+          ConstrainableDataTypes.Numeric
+        }
+      case DataTypeInstances.Integral =>
+        if (isTypeEmpty(DataTypeInstances.Fractional)) {
+          ConstrainableDataTypes.Integral
+        } else {
+          ConstrainableDataTypes.Numeric
+        }
       case DataTypeInstances.Boolean => ConstrainableDataTypes.Boolean
     }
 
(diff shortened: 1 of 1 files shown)


sscdotopen:
I think this is based on a misunderstanding. The idea of the type detection is to detect to which type a particular attribute can be safely cast. If a column has both fractional and integer values, the `Fractional` answer is the correct one, as it can be safely cast to fractional values.

aviatesk:
Hmm, so you mean when a check with  suggested `.hasDataType(column, Fractional)` fails, we're supposed to cast the `column` into `Fractional` type ahead of deequ check, right ?

My original idea was that suggested checks are better to not fail on the original data from which they're generated.
But this is somewhat against how we're supposed to use this feature ?

sscdotopen:
> Hmm, so you mean when a check with suggested `.hasDataType(column, Fractional)` fails, we're supposed to cast the `column` into `Fractional` type ahead of deequ check, right ?

Yes, exactly

> My original idea was that suggested checks are better to not fail on the original data from which they're generated.
> But this is somewhat against how we're supposed to use this feature ?

That's a good point, maybe we should extend the checks to include the cast? What do you think

aviatesk:
> That's a good point, maybe we should extend the checks to include the cast? What do you think

As far as I understand how `.hasDataType(column, Numeric)` works, it's equivalent to casting, no ? (assuming we can always cast `Integral` values into `Fractional` ones)
https://github.com/awslabs/deequ/blob/0f26ea5d5acba4ffa2022d7653032832b822b551/src/main/scala/com/amazon/deequ/constraints/Constraint.scala#L618-L632

And I guess the current design may not allow us to include casting in `.hasDataType` check, because it's implemented as `AnalysisBasedConstraint`, which works on already-calculated metrics.

sscdotopen:
hasDataType doesn't need to cast, it checks string values similar to the profiling code. I think the problematic cases are checks like `isNonNegative`, which require a cast.

aviatesk:
> hasDataType doesn't need to cast

No, with the following MRE, we get (opinionatedly unexpected) anomaly reports from `.hasDataType`:
```scala
import com.amazon.deequ.{VerificationSuite, VerificationResult}
import com.amazon.deequ.checks.{Check, CheckLevel, CheckStatus}
import com.amazon.deequ.constraints.ConstraintStatus
import com.amazon.deequ.constraints.{ConstrainableDataTypes}

case class Item(v: String)

val rdd = spark.sparkContext.parallelize(Seq(
  Item("1.0"),
  Item("2.0"),
  Item("3"),
  Item("4.0")
))
val data = spark.createDataFrame(rdd)

val result = VerificationSuite()
  .onData(data)
  .addCheck(
    Check(CheckLevel.Error, "I want this to pass")
      .hasDataType("v", ConstrainableDataTypes.Fractional)
  )
  .run()

VerificationResult.checkResultsAsDataFrame(spark, result).show()
```
```
```

twollnik:
Hi, thanks for the suggestion and the discussion so far. We discussed this in the team and would prefer a different approach:

- parameterize the `.hasDataType` check to distinguish between strict and non-strict mode. Non-strict mode could use casting internally to not fail in the case that you outlined above.
- the suggestion could suggest strict vs non-strict mode depending on whether or not doubles and ints are mixed in the input.

Please let us know what you think about this and if you think you can find the time to make this change. Thanks!

twollnik:
@aviatesk @sscdotopen friendly ping

twollnik:
Closing due to inactivity. Please re-open if this is still a concern. Thanks for your contribution!

PR #332, opened 2020-12-28
sort suggested categories

aviatesk:
*Issue #, if available:*

*Description of changes:*

- sort suggested categorical values when generating constraint code (sort in lexical order)
- this may improve the readability of suggested constraints and help one refer to it as some form of the data schema

### DIFF
--- a/src/main/scala/com/amazon/deequ/suggestions/rules/CategoricalRangeRule.scala
+++ b/src/main/scala/com/amazon/deequ/suggestions/rules/CategoricalRangeRule.scala
@@ -22,9 +22,13 @@ import com.amazon.deequ.constraints.Constraint.complianceConstraint
 import com.amazon.deequ.profiles.ColumnProfile
 import com.amazon.deequ.suggestions.ConstraintSuggestion
 import org.apache.commons.lang3.StringEscapeUtils
+import com.amazon.deequ.metrics.DistributionValue
 
 /** If we see a categorical range for a column, we suggest an IS IN (...) constraint */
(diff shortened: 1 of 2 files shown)


aviatesk:
@twollnik can we push this forward ?
This PR doesn't change any suggestion functionality itself, but just improve its referability, e.g.

> before
```scala
// suggested
isContainedIn("area_cd", Array("030", "060", "050", "090", "020", "080", "040", "010", "070"))
```

> after
```scala
// suggested
.isContainedIn("area_cd", Array("010", "020", "030", "040", "050", "060", "070", "080", "090")),
```

twollnik:
We are explicitely ordering the values by popularity just a couple of lines earlier (`.sortBy { case (_, value) => value.absolute }`). @sscdotopen Do you know if there is a specific reason for this?

aviatesk:
Ops, I just assumed they were unsorted.
Then maybe we should make it configurable at least ? Something like:
```scala
case class CategoricalRangeRule[A](
  sorter: ((String, DistributionValue)) => A = (a: (String, DistributionValue)) => a match { case (_, value) => value.absolute }
)(implicit ord: Ordering[A]) extends ConstraintRule[ColumnProfile] {
```

aviatesk:
@twollnik any update ? How about adding `sorter` constructor argument ? An user can opt into to change that, but the default behaviour won't change.

aviatesk:
okay, this should be good to go.
I decided to add a constructor option that allows each user to give their own category sorter. Since the default value is preserved same as the previous implementation, this PR shouldn't change any behavior by itself.

twollnik:
Thanks for your patience on this!

PR #338, opened 2021-01-29
separate constraint hint from constraint message

aviatesk:
*Issue #, if available:*

- currently constraint hint is truncated into constraint message
- when we want to use it to making alert etc., I think it would be nicer if they're separated because
  * while constraint message provides "why this constraint failed",
  * constraint hint should provide "why this constraint can fail"
- and we may want to use them for separate purposes.

*Description of changes:*

separate constraint hint from constraint message

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #339, opened 2021-02-03
include hints for suggested constraints

aviatesk:
*Issue #, if available:*

*Description of changes:*

- built on top of #338 
- includes hints for suggested constraints, which should vastly increase interpretability


twollnik:
Hi @aviatesk, thanks for suggesting this change. We don't yet have a good enough understanding of your use-case to feel confident in making this change. Could you please explain some more why this is necessary and how you are using the hints? Also, how did you arrive at the `Some("..")` format for hints?

twollnik:
@aviatesk please get back to us on this if you get the chance. We are considering closing this PR soon.

PR #342, opened 2021-02-12
allow nullable matcher pattern for `hasPattern`

aviatesk:
*Issue #, if available:*

Currently `.hasPattern` doesn't allow a pattern that matches `null`s  to work,
because our aggregation logic checks for the matched value from `regexp_extract` is `""` or not.

*Description of changes:*

We will use `.rlike` instead of `regexp_extract`.


src/test/scala/com/amazon/deequ/checks/CheckTest.scala
```diff
@@ -626,6 +626,24 @@ class CheckTest extends AnyWordSpec with Matchers with SparkContextSpec with Fix
...
 
+    "allow regular expression pattern to match for null values" in withSparkSession { spark =>
+      import spark.implicits._
+
+      // XXX `id` is needed, otherwise scalastyle will complain due to the bug below
+      // https://github.com/scalastyle/scalastyle/issues/195
+      val df = Seq(
+        ("123", 1),
+        (null, 2)
+      ).toDF("nullable", "id")
```
aviatesk, line 637: `scalastyle` isn't too interesting, imho.
Can we remove the integration ?

twollnik:
Hi, thanks for introducing these changes. We don't feel comfortable changing the behaviour of this check (e.g. interpreting empty string as successful match). We could move forward with this if we introduce a parameter that enables users to switch between the two modes (defaulting to the old behaviour).

twollnik:
@aviatesk please get back to us on this if you get the chance. We are considering closing this PR soon.

twollnik:
Closing due to inactivity. If this is still a concern, please feel free to re-open.

PR #370, opened 2021-06-12
Fix PatternMatch hashCode bug

zhuowang-linkedin:
*Issue #, if available:*

*Description of changes:*
The PatternMatch hashCode differs for the same class parameters because Regex parameter compares by reference other than the pattern value.
A possible problem this causing is getting PatternMatch metrics from AnalyzerContext map will return nothing if the PatternMatch object does not reference equal.


zhuowang-linkedin:
ping

twollnik (approved):
Looks good, thanks for introducing these changes.

Issue #371, opened 2021-06-17
Improve usability of verification suite results if validation is a success


andreaslang:
I am not sure if I am missing something, but I have been struggling making historical verification suite results usable in the sense of that I know what the actual condition was. I have been using hints, which makes things better for failed checks (I actually see what I was checking for), but hints are not added if the assertion was OK.

This code in AnalysisBasedConstraint.scala is what I mean in particular.
```scala
          if (assertionOk) {
            ConstraintResult(this, ConstraintStatus.Success, metric = Some(metric))
          } else {
            var errorMessage = s"Value: $assertOn does not meet the constraint requirement!"
            hint.foreach(hint => errorMessage += s" $hint")

            ConstraintResult(this, ConstraintStatus.Failure, Some(errorMessage), Some(metric))
          }
```

I feel like I am missing a trick here, maybe also caused by the fact that I am using things indirectly via pydeequ, but isn't there a way to get to some output, which tells me: You checked for this (via hint), the metric was this (via assertOn) and all was fine.

Essentially this would be perfect:
```scala
          if (assertionOk) {
            val successMessage = s"Value: $assertOn does meet the constraint requirement!"
            hint.foreach(hint => successMessage += s" $hint")

            ConstraintResult(this, ConstraintStatus.Success, Some(successMessage), Some(metric))
          } else {
            var errorMessage = s"Value: $assertOn does not meet the constraint requirement!"
            hint.foreach(hint => errorMessage += s" $hint")

            ConstraintResult(this, ConstraintStatus.Failure, Some(errorMessage), Some(metric))
          }
```

twollnik:
We like this idea, can you submit a PR?

andreaslang:
Sure can do, will put something up in the next few days.

twollnik:
Thanks. Don't worry about doing this quickly, we likely won't get around to reviewing the PR until mid august or end of august anyway.

PR #375, opened 2021-08-06
Display success message in Verification Suite Result

andreaslang:
https://github.com/awslabs/deequ/issues/371

Display also a success message for verification, seeing the hint and assertOn value makes it easier to see which verification succeeded.

### DIFF
--- a/src/test/scala/com/amazon/deequ/VerificationResultTest.scala
+++ b/src/test/scala/com/amazon/deequ/VerificationResultTest.scala
@@ -124,11 +124,11 @@ class VerificationResultTest extends WordSpec with Matchers with SparkContextSpe
         import session.implicits._
         val expected = Seq(
           ("group-1", "Error", "Success", "CompletenessConstraint(Completeness(att1,None))",
-            "Success", "Value: 1.0 does meet the constraint requirement!"),
+            "Success", "Value: 1.0 meets the constraint requirement."),
           ("group-2-E", "Error", "Error", "SizeConstraint(Size(None))", "Failure",
             "Value: 4 does not meet the constraint requirement! Should be greater than 5!"),
           ("group-2-E", "Error", "Error", "CompletenessConstraint(Completeness(att2,None))",
-            "Success", "Value: 1.0 does meet the constraint requirement! Should equal 1!"),
+            "Success", "Value: 1.0 meets the constraint requirement. Should equal 1!"),
           ("group-2-W", "Warning", "Warning",
             "DistinctnessConstraint(Distinctness(List(item),None))",
             "Failure", "Value: 1.0 does not meet the constraint requirement! " +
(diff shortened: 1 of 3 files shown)


src/test/scala/com/amazon/deequ/VerificationResultTest.scala
```diff
@@ -124,11 +124,11 @@ class VerificationResultTest extends WordSpec with Matchers with SparkContextSpe
...
         val expected = Seq(
           ("group-1", "Error", "Success", "CompletenessConstraint(Completeness(att1,None))",
-            "Success", ""),
+            "Success", "Value: 1.0 does meet the constraint requirement!"),
           ("group-2-E", "Error", "Error", "SizeConstraint(Size(None))", "Failure",
             "Value: 4 does not meet the constraint requirement! Should be greater than 5!"),
           ("group-2-E", "Error", "Error", "CompletenessConstraint(Completeness(att2,None))",
-            "Success", ""),
+            "Success", "Value: 1.0 does meet the constraint requirement! Should equal 1!"),
           ("group-2-W", "Warning", "Warning",
```
twollnik, line 132: Can you clarify why the message contains the hint ("Should ...!") in one case, but not in the other?

twollnik (commented):
Thanks for this contribution, we appreciate the changes. Just one clarification needed.

src/test/scala/com/amazon/deequ/VerificationResultTest.scala
andreaslang: Sure, the reason is that the first check does not actually have a hint. In that case only the standard message is shown.

The checks are in the same test suite from line 201:

```scala
private[this] def getChecks(): Seq[Check] = {
    val checkToSucceed = Check(CheckLevel.Error, "group-1")
      .isComplete("att1")

    val checkToErrorOut = Check(CheckLevel.Error, "group-2-E")
      .hasSize(_ > 5, Some("Should be greater than 5!"))
      .hasCompleteness("att2", _ == 1.0, Some("Should equal 1!"))

    val checkToWarn = Check(CheckLevel.Warning, "group-2-W")
      .hasDistinctness(Seq("item"), _ < 0.8, Some("Should be smaller than 0.8!"))

    checkToSucceed :: checkToErrorOut :: checkToWarn :: Nil
}
```
  
The change allows the hint to be passed through even if the evaluation was a success, before that the hint was lost. Passing it through makes it a bit easier if the verification result is saved, because the hint can be used to know what the threshold at a given time was.

twollnik:
Thanks for clarification. One last request: can we change the message for the success case to `"... meets the constraint requirement."`? We think this would make it easier to distinguish between the success messages and the error messages.

andreaslang:
Sorry for the delay, was moving house. Makes sense, I updated the message.

andreaslang:
> Thanks for clarification. One last request: can we change the message for the success case to `"... meets the constraint requirement."`? We think this would make it easier to distinguish between the success messages and the error messages.

Hi @twollnik , given that change is done, just wanted to check if you need anything else. Apologies again for providing the change for that a bit late.

Issue #271, opened 2020-08-04
MetricRepository cannot store metrics of Histogram analyzer with filter


pwzhong:
**Bug: MetricRepository cannot store metrics of Histogram analyzer with filter correctly. It is stored as Histogram metric without any filter.**

Here is a small snippet to demo the issue, using latest **build 1.0.4**:
```
    val path = <repository_path>
    val spark: SparkSession = ...

    import spark.implicits._
    val inputDF = Seq(("a", 1),("a", 1),("a", 2),("a", 3),("b", 1),("b", 2),("b", 2),("c",1)).toDF("id", "value")

    val repository = FileSystemMetricsRepository(spark, path)
    val resultKey = ResultKey(System.currentTimeMillis(), Map("tag" -> "test"))

    // collect Histogram metrics with filter and store in the repository
    val analysisResult = AnalysisRunner
      .onData(inputDF)
      .useRepository(repository)
      .addAnalyzers(Seq(Histogram("value",where=Some("id='a'")), Histogram("value",where=Some("id='b'")), Histogram("value",where=Some("id='c'"))))
      .saveOrAppendResult(resultKey)
      .run()

    // print out the collected metric. It shows Histogram metrics with filter are collected correctly.
    AnalyzerContext.successMetricsAsDataFrame(spark, analysisResult).show(false)

   // print out the collected metric loading from the repository. Here is the error: the filter is missing. It is stored as Histogram metric without any filter.
    println("Data stored in metric repository: ")
    repository.load()
    .withTagValues(Map("tag" -> "test"))
    .getSuccessMetricsAsDataFrame(spark)
    .show(false)
```
**Result:**
```
+------+--------+---------------------------------+-----+
|entity|instance|name                             |value|
+------+--------+---------------------------------+-----+
|Column|value   |Histogram.bins (where: id='a')   |3.0  |
|Column|value   |Histogram.abs.1 (where: id='a')  |2.0  |
|Column|value   |Histogram.ratio.1 (where: id='a')|0.25 |
|Column|value   |Histogram.abs.3 (where: id='a')  |1.0  |
|Column|value   |Histogram.ratio.3 (where: id='a')|0.125|
|Column|value   |Histogram.abs.2 (where: id='a')  |1.0  |
|Column|value   |Histogram.ratio.2 (where: id='a')|0.125|
|Column|value   |Histogram.bins (where: id='b')   |2.0  |
|Column|value   |Histogram.abs.2 (where: id='b')  |2.0  |
[...]

PR #379, opened 2021-08-27
fixed issue #271: Histogram analyzer filter missing

pwzhong:
*Issue #, if available: #271

*Description of changes:*
Added the missing filter to Histogram analyzer in AnalysisResultSerde.scala, so that Histogram metrics with filter can be stored in the repository.

P.S. The issue was opened last year and unfortunately I haven't got a chance to submit a PR. Recently, I checked the latest build that this bug still exists, so I submit it this time. I was also thinking about if a test case is needed. In AnalysisResultSerdeTest.scala, it seems nowhere to fit in such a use case. Please let me know if any suggestion. Thanks.


lange-labs:
Thanks for submitting the change. Could you check if you can add a test case around here:
https://github.com/awslabs/deequ/blob/32be67a4b4f9c02a139f7214f06621490ffd884d/src/test/scala/com/amazon/deequ/repository/AnalysisResultSerdeTest.scala#L64

sudsali:
Closing since the issue has been resolved in this #708 PR.

PR #386, opened 2021-09-30
Make serde classes public

jameskyle:
*Description of changes:*

This PR lifts the private restrictions on the SERDE classes. It also reorganizes the code to a single object/class per file for easier code navigation and discovery.

No functional or dependency changes were made.

My personal motivation for these changes is the ability to have an entirely configuration driven validation class where checks and metrics are defined in a json, yaml, or hocon configuration file.

However, I see it's come up a couple of times. Like the PR for [making the AnalysisResult SERDE public](https://github.com/awslabs/deequ/pull/117/files) and the Extending MetricsRepository Issue #232.

### DIFF
--- a/src/main/scala/com/amazon/deequ/repository/AnalysisResultDeserializer.scala
+++ b/src/main/scala/com/amazon/deequ/repository/AnalysisResultDeserializer.scala
@@ -1,0 +1,50 @@
+/** Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+  *
+  * Licensed under the Apache License, Version 2.0 (the "License"). You may not
+  * use this file except in compliance with the License. A copy of the License
+  * is located at
+  *
+  *     http://aws.amazon.com/apache2.0/
+  *
+  * or in the "license" file accompanying this file. This file is distributed on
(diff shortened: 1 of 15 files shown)


jameskyle:
I still think this is valuable, but in my own code I ended up using Lift's native scala json library. I'd actually recommend migrating this code over to that library for a few reasons:

* Avoids the necessity to move types into/out of native java objects.
* Does not require an explicit serializer. Field names map to method names for case classes, for example. 
* much, much less boiler plate. 

For example, here's a serialized histogram

```
    histogram: [
      {
        "column": "state"
        "maxDetailBins": 1000
        "where": "iso_country_code = 'US'"
      }
```

And here's how you deserialize it

```
import net.liftweb.json.{DefaultFormats, JValue, parse}
import com.amazon.deequ.analyzers.Histogram

val check = parse(jsonString).extract[Histogram]
```

I didn't dive in and go this route because the current Serde design is effected to get the cleanest implementation. So it would take a bit more thought to migrate.

buddiex:
Any plans to merge this?
@jameskyle @twollnik

jameskyle:
> Any plans to merge this? @jameskyle @twollnik

I actually took a different approach. writing a serde that did not require juggling java to scala types.

If this is still seen as valuable, I can merge it.

PR #396, opened 2021-11-08
Cast column types with select() instead of withColumn()

rli602:
*Issue #, if available:* N/A

*Description of changes:*
We're facing slowness while running the ColumnProfiler on large datasets. According to the PySpark's [doc](https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.sql.DataFrame.withColumn.html), using withColumn for multiple times, like in a for-loop, can cause performance issues. select() is suggested here to replace the multiple withColumn() call.


sudsali:
Closing because the fix (#711) has been merged.

PR #436, opened 2022-10-04
Referential/2.0.0 spark 3.1

dariobig:
*Issue #, if available:*

*Description of changes:*

New Referential integrity and Data synchronization checks

### DIFF
--- a/src/main/scala/com/amazon/deequ/comparison/DataSynchronization.scala
+++ b/src/main/scala/com/amazon/deequ/comparison/DataSynchronization.scala
@@ -19,84 +19,111 @@ import org.apache.spark.sql.functions.col
 
 object DataSynchronization {
 
-  /** @param ds1:
+  /** @param df1:
     *   The dataset from which the keys of the mergedMap are retrieved from.
-    * @param ds2:
+    * @param df2:
     *   The dataset from which the values of the mergedMap are retrieved from.
     * @param mergedMaps:
-    *   A map that contains the columns that the customer is going to compare.
+    *   A map that contains the columns that the user is going to compare.
     * @return
     *   Double: Returns the amount of rows of the joined DataFrame as a Double
     */
-
   def mergingMaps(
-      ds1: DataFrame,
-      ds2: DataFrame,
+      df1: DataFrame,
+      df2: DataFrame,
       mergedMaps: Map[String, String]
   ): Double = {
     val joinExpression: Column = mergedMaps
       .map { case (col1, col2) =>
-        ds1(col1) === ds2(col2)
+        df1(col1) === df2(col2)
       }
       .reduce((e1, e2) => e1 && e2)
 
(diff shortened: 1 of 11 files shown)


src/main/scala/com/amazon/deequ/comparison/ReferentialIntegrity.scala
```diff
@@ -0,0 +1,44 @@
...
+object ReferentialIntegrity {
+
+  def subsetCheck(
+      ds1: DataFrame,
+      col1: String,
+      ds2: DataFrame,
+      col2: String,
+      assertion: Double => Boolean
+  ): Boolean = {
+
```
eycho-am, line 28: Should we add documentation for the method and params?

eycho-am (commented):
All the tests pass when I pulled the code, just left a comment for missing documentation.

src/main/scala/com/amazon/deequ/comparison/ReferentialIntegrity.scala
dariobig: Generally code should speak for itself, but I agree with you here.

dariobig: fixed

src/main/scala/com/amazon/deequ/comparison/DataSynchronization.scala
```diff
@@ -0,0 +1,131 @@
...
+    */
+  def columnMatch(
+      df1: DataFrame,
+      df2: DataFrame,
+      primaryKeyColumns: Map[String, String],
+      joinColumns: Option[Map[String, String]],
+      assertion: Double => Boolean
+  ): Boolean = {
+
+    val df1Unique =
```
dariobig, line 83: I suspect this can be optimized, especially the custom join bit, we shuold setup a benchmark for it

dariobig: TODO: needs an issue

src/main/scala/com/amazon/deequ/comparison/DataSynchronization.scala
```diff
@@ -0,0 +1,131 @@
...
+      mergedMaps: Map[String, String]
+  ): Double = {
+    val joinExpression: Column = mergedMaps
+      .map { case (col1, col2) =>
+        df1(col1) === df2(col2)
+      }
+      .reduce((e1, e2) => e1 && e2)
+
+    // joining the df1 with df2 with only the rows that matches obtained from joinExpression
+    val joined = df1.join(df2, joinExpression, "inner")
```
dariobig, line 43: see below, I suspect doing a normal join and then figuring out how many rows were lost is faster

dariobig: maybe even using a semi-join + caching

PR #451, opened 2023-02-09
Feature: Row Level Results

mentekid:
*Issue #, if available:*
N/A

*Description of changes:*

This is an early version of the row-level results feature. I have enabled row-level results in three constraints: `IsComplete`, `HasCompleteness`, and `MaxLength`.

This requires the definition and change of certain classes defined in Deequ both to expose information previously not available, and to allow distinguishing between different analyzers' results.

### DIFF
--- a/src/main/scala/com/amazon/deequ/VerificationResult.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationResult.scala
@@ -86,12 +86,9 @@ object VerificationResult {
       verificationResult: VerificationResult,
       data: DataFrame): DataFrame = {
 
-    val columnNamesToMetrics: Map[String, Metric[_]] = mapResultToMetric(verificationResult)
-      .flatMap(mapColumnNameToMetric)
+    val columnNamesToMetrics: Map[String, Column] = verificationResultToColumn(verificationResult)
 
-    val checks: Seq[(String, Column)] = columnNamesToMetrics.flatMap(getColumnIfCalculated).toList
-
-    checks.foldLeft(data)(
+    columnNamesToMetrics.foldLeft(data)(
       (data, newColumn: (String, Column)) => data.withColumn(newColumn._1, newColumn._2))
   }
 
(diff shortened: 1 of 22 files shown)


src/main/scala/com/amazon/deequ/VerificationResult.scala
```diff
@@ -70,6 +77,23 @@ object VerificationResult {
       "constraint_status", "constraint_message")
   }
 
+  /**
+   * For each check in the verification suite, adds a column of row-level results
+   * to the input data if that check contains a column.
+   *
+   * Accepts a naming rule
+   */
+  def toRowLevelResults(
```
rdsharma26, line 86: Minor: Can we name this in accordance with the other public methods? `rowLevelResultsAsDataFrame` perhaps?

src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
```diff
@@ -41,4 +41,6 @@ case class MaxLength(column: String, where: Option[String] = None)
   }
 
   override def filterCondition: Option[String] = where
+
+  private def criterion: Column = length(conditionalSelection(column, where)).cast(DoubleType)
```
rdsharma26, line 45: This is a private method here and a public method in Completeness. If it is required in each analyzer, should it be part of a base class?

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -414,7 +433,13 @@ object Constraint {
     val constraint = AnalysisBasedConstraint[MaxState, Double, Double](maxLength, assertion,
       hint = hint)
 
-    new NamedConstraint(constraint, s"MaxLengthConstraint($maxLength)")
+    val sparkAssertion = org.apache.spark.sql.functions.udf(assertion)
+
+    new RowLevelAssertedConstraint(
+      constraint,
+      s"MaxLengthConstraint($maxLength)",
+      s"StringLength-$column",
```
rdsharma26, line 441: Minor: Would `ColumnLength` be a better name? It remains consistent with the terminology in Glue. `StringLength` is a new term.

src/main/scala/com/amazon/deequ/metrics/Metric.scala
```diff
@@ -37,13 +39,24 @@ trait Metric[T] {
...
 }
 
+/**
+ * Full-column metrics store the entire column of row-level pass/fail results
+ */
+trait FullColumn {
+  val fullColumn: Option[Column] = None
+
+  def sum(colA: Option[Column], colB: Option[Column]): Option[Column] =
+    if (colA.equals(colB)) colA else None
```
rdsharma26, line 49: The implementation of this function does not seem intuitive. Can we add comments what the function is doing?

src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
mentekid: the other one is used by a test that constructs an expected object and compares the result of Completeness to that object. It shouldn't be public because I don't think anything in the code should invoke it. I'll make it package-private inside deequ and mark it as `@VisibleForTesting` to point out it's an internal detail for the analyzer

mentekid:
Opening this against master instead https://github.com/awslabs/deequ/pull/452

rdsharma26:
Thanks for updating the branch to master. LGTM.

rdsharma26 (approved):

PR #453, opened 2023-03-01
Fix style issues causing mvn install to fail.

rdsharma26:
*Description of changes:*

- Style issues were caused due to spacing issues and the presence of the chi character in unicode.
- The workflow step was also updated to include the execution of the style check. Previously, the style check was not being executed which led to the style errors being committed to master.


bevhanno (approved):
approved

eycho-am (approved):
lgtm

mentekid (approved):

Issue #274, opened 2020-08-09
Problem with colum names that have dot charachter 


dmiljkovic:
Below constraint suggestion code produces 

org.apache.spark.sql.AnalysisException: syntax error in attribute name: Phone No.;

The problem is that dot character is a part of the column name  (Phone No.). When I remove dot from name code runs fine. 

import com.amazon.deequ.SparkContextSpec
import com.amazon.deequ.suggestions.{ConstraintSuggestionRunner, Rules}
import com.amazon.deequ.utils.FixtureSupport
import org.apache.spark.sql.{DataFrame, SparkSession}
import org.scalatest.{Matchers, WordSpec}

class TestSuggestions extends WordSpec with Matchers with SparkContextSpec with FixtureSupport {

  def testData(sparkSession: SparkSession): DataFrame = {
    import sparkSession.implicits._

    Seq(
      ("CA", "123"),
      ("SD", "1233"),
      ("NC", "1236")
    )
      .toDF("state", "Phone No.")
  }

  "column name problem" should {

    "" in withSparkSession { session =>

      val data = testData(session)
      data.show(false)
      val suggestionResult = ConstraintSuggestionRunner()
        .onData(data)
        .addConstraintRules(Rules.DEFAULT)
        .run()
      suggestionResult.constraintSuggestions.foreach { case (column, suggestions) =>
        suggestions.foreach { suggestion =>
          println(s"Constraint suggestion for '$column':\t${suggestion.description}\n" +
            s"The corresponding scala code is ${suggestion.codeForConstraint}\n")
        }
      }
    }
  }
}

sscdotopen:
This is a known bug unfortunately, we do not correctly escape the column names in all places. We had a person working on this, but they unfortunately never finished their PR. As a work around, you could rename the column on the dataframe before running the test.

TianLangStudio:
> This is a known bug unfortunately, we do not correctly escape the column names in all places. We had a person working on this, but they unfortunately never finished their PR. As a work around, you could rename the column on the dataframe before running the test.

Hello, Somebody is working on this?  I want to try to fix this.   I saw this project in Upwork and very interesting in it.

sscdotopen:
Feel free to submit a PR for that, we would be very happy. Btw, what exactly do you mean by "saw this project on upwork"?

TianLangStudio:
Thanks, I'll try to do it.  I just say how I find this project. it's don't matter.  
[...]

PR #461, opened 2023-03-28
Bugfix/escape columnname

eycho-am:
*Issue #, if available:* [#274](https://github.com/awslabs/deequ/issues/274)

*Description of changes:*

When there is a . in the column name, Spark recognizes that as trying to get the field nested in a column. This is an issue for those testing with common data sets like Iris, resulting in the following error: 
`org.apache.spark.sql.AnalysisException: Column 'sepal.width' does not exist. Did you mean one of the following? [petal.length, petal.width, sepal.length, sepal.width, variety]`

This PR addresses the issue by using helper methods that add and remove backticks "`" in the appropriate instances when going through the data.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -204,7 +204,6 @@ abstract class StandardScanShareableAnalyzer[S <: DoubleValuedState[_]](
     entity: Entity.Value = Entity.Column)
   extends ScanShareableAnalyzer[S, DoubleMetric] {
 
-  // Thrown here
   override def computeMetricFrom(state: Option[S]): DoubleMetric = {
     state match {
       case Some(theState) =>
@@ -218,7 +217,6 @@ abstract class StandardScanShareableAnalyzer[S <: DoubleValuedState[_]](
     }
   }
 
-  // Thrown here
   override private[deequ] def toFailureMetric(exception: Exception): DoubleMetric = {
     metricFromFailure(exception, name, instance, entity)
   }
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -23,6 +23,7 @@ import org.apache.spark.sql.types._
 import org.apache.spark.sql.{Column, DataFrame, Row, SparkSession}
 import com.amazon.deequ.analyzers.runners._
 import com.amazon.deequ.metrics.FullColumn
+import com.amazon.deequ.utilities.ColumnUtil.removeEscapeColumn
 
 import scala.language.existentials
 import scala.util.{Failure, Success}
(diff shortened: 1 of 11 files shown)


src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -204,6 +204,7 @@ abstract class StandardScanShareableAnalyzer[S <: DoubleValuedState[_]](
     entity: Entity.Value = Entity.Column)
   extends ScanShareableAnalyzer[S, DoubleMetric] {
 
+  // Thrown here
```
rdsharma26, line 207: nit: Remove comment or rephrase it with more details

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -217,6 +218,7 @@ abstract class StandardScanShareableAnalyzer[S <: DoubleValuedState[_]](
     }
   }
 
+  // Thrown here
```
rdsharma26, line 221: nit: Remove comment or rephrase it with more details

src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
```diff
@@ -299,6 +299,7 @@ object AnalysisRunner {
     frequenciesAndNumRows.numRows -> results
   }
 
+  // Main method that needs looking into
```
rdsharma26, line 302: nit: Remove comment or rephrase it with more details

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -303,17 +305,34 @@ object Preconditions {
...
+  }
+
+  def escapeColumn(column: String): String = {
+    if (column.contains(".")) {
+      "`" + column + "`"
+    } else {
+      column
+    }
+  }
+
```
rdsharma26, line 329: Can we create a utility object which contains these methods? In the future, one wouldn't immediately know that the important methods for escaping/un-escaping are inside Analyzer.scala

src/main/scala/com/amazon/deequ/profiles/ColumnProfiler.scala
```diff
@@ -450,15 +471,13 @@ object ColumnProfiler {
...
-
       castedData = genericStatistics.typeOf(name) match {
         case Integral => castColumn(castedData, name, LongType)
         case Fractional => castColumn(castedData, name, DoubleType)
         case _ => castedData
       }
     }
-
-    castedData
+  castedData
```
rdsharma26, line 480: nit: inconsistent indentation?

rdsharma26 (commented):
Great PR to fix a long-standing issue!

eycho-am:
> When you say that Spark recognizes `.` as trying to get the field nested in a column, does that mean Deequ can work with nested fields? Will `.` work to add constraints on nested fields? If so, can we add a test for this behavior as a means of documentation.

I tried creating a test for this but based on this method it looks like Deequ doesn't work with nested columns:
https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala#L354

rdsharma26 (approved):

PR #465, opened 2023-04-11
Feature: Length Row Level Results

eycho-am:
*Issue #, if available:*
N/A

*Description of changes:*
Building on this [PR](https://github.com/awslabs/deequ/pull/451) to add more analyzers that support row level results.

This PR address this for `MinLength`, as `MaxLength` was completed in the PR mentioned above.

In addition, this PR adds an `AnalzyerOptions` case class that gives an option to convert null values to result to failing the check. This option is not set by default, which means that for a `MinLength` analyzer, the check would succeed even if null values were present.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -17,16 +17,22 @@
 package com.amazon.deequ.analyzers
 
 import com.amazon.deequ.analyzers.Analyzers._
-import com.amazon.deequ.metrics.{DoubleMetric, Entity, Metric}
-import org.apache.spark.sql.functions._
-import org.apache.spark.sql.types._
-import org.apache.spark.sql.{Column, DataFrame, Row, SparkSession}
 import com.amazon.deequ.analyzers.runners._
 import com.amazon.deequ.metrics.FullColumn
 import com.amazon.deequ.utilities.ColumnUtil.removeEscapeColumn
+import com.amazon.deequ.metrics.DoubleMetric
+import com.amazon.deequ.metrics.Entity
+import com.amazon.deequ.metrics.Metric
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.SparkSession
 
 import scala.language.existentials
-import scala.util.{Failure, Success}
+import scala.util.Failure
+import scala.util.Success
 
 /**
   * A state (sufficient statistic) computed from data, from which we can compute a metric.
@@ -449,6 +455,12 @@ private[deequ] object Analyzers {
     if (columns.size == 1) Entity.Column else Entity.Mutlicolumn
   }
 
+  def conditionalSelectionForLength(selection: Column, where: Option[String], replaceWith: Double): Column = {
+    val conditionColumn = where.map { expression => expr(expression) }
+    conditionColumn
+      .map { condition => when(condition, replaceWith).otherwise(selection) }
+      .getOrElse(selection)
+  }
   def conditionalSelection(selection: String, where: Option[String]): Column = {
     conditionalSelection(col(selection), where)
   }
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -255,6 +255,12 @@ case class NumMatchesAndCount(numMatches: Long, count: Long, override val fullCo
   }
 }
 
+case class AnalyzerOptions(convertNull: Boolean) {
+  def getConvertNull(): Boolean = {
+    convertNull
+  }
+}
+
 /** Base class for analyzers that compute ratios of matching predicates */
 abstract class PredicateMatchingAnalyzer(
     name: String,
--- a/src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
@@ -26,7 +26,7 @@ import org.apache.spark.sql.functions.max
 import org.apache.spark.sql.types.DoubleType
 import org.apache.spark.sql.types.StructType
 
-case class MaxLength(column: String, where: Option[String] = None, convertNull: Boolean = false)
+case class MaxLength(column: String, where: Option[String] = None, analyzerOptions: Option[AnalyzerOptions] = None)
   extends StandardScanShareableAnalyzer[MaxState]("MaxLength", column)
   with FilterableAnalyzer {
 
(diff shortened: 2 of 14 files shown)


src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -249,6 +255,12 @@ case class NumMatchesAndCount(numMatches: Long, count: Long, override val fullCo
   }
 }
 
+case class AnalyzerOptions(convertNull: Boolean) {
+  def getConvertNull(): Boolean = {
```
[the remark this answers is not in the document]
eycho-am: Changed it to use an enum `NullBehavior` that gives users 3 options
1. Ignore (Default behavior)
2. Empty (Treat as 0 or empty string)
3. Fail (Always fail nulls)

src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
```diff
@@ -42,5 +49,13 @@ case class MaxLength(column: String, where: Option[String] = None)
 
   override def filterCondition: Option[String] = where
 
-  private def criterion: Column = length(conditionalSelection(column, where)).cast(DoubleType)
+  private def criterion(convertNull: Boolean): Column = {
+    if (convertNull) {
+      val colLengths: Column = length(conditionalSelection(column, where)).cast(DoubleType)
+      conditionalSelectionForLength(colLengths, Option(s"${column} IS NULL"), Double.MinValue)
```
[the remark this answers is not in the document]
eycho-am: Now with the `NullBehavior` enum, users can choose to replace with "" or treat as false (fixed to Double.MaxValue).

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -453,6 +471,13 @@ private[deequ] object Analyzers {
     conditionalSelection(col(selection), where)
   }
 
+  def conditionalSelection(selection: Column, where: Option[String], replaceWith: Any): Column = {
```
eycho-am, line 474: I made `replaceWith` as type `Any` so different types can be replaced depending on the analyzers. Would it be better to leave it as Double for the length analyzers?

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -453,6 +471,20 @@ private[deequ] object Analyzers {
     conditionalSelection(col(selection), where)
   }
 
+  def conditionalSelection(selection: Column, where: Option[String], replaceWith: Double): Column = {
+    val conditionColumn = where.map { expression => expr(expression) }
```
rdsharma26, line 475: nit: `where.map(expr)`

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -453,6 +471,20 @@ private[deequ] object Analyzers {
...
+    conditionColumn
+      .map { condition => when(condition, replaceWith).otherwise(selection) }
+      .getOrElse(selection)
+  }
+
+  def conditionalSelection(selection: Column, where: Option[String], replaceWith: String): Column = {
+    val conditionColumn = where.map { expression => expr(expression) }
```
rdsharma26, line 482: nit: `where.map(expr)`

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -249,6 +256,17 @@ case class NumMatchesAndCount(numMatches: Long, count: Long, override val fullCo
   }
 }
 
+case class AnalyzerOptions(nullBehavior: NullBehavior = NullBehavior.Ignore) {
+  def getNullBehavior(): NullBehavior = {
+    nullBehavior
+  }
```
rdsharma26, line 262: Not required.
```
scala> case class Options(iLikeNulls: Boolean)
defined class Options

scala> val opt = Options(iLikeNulls = false)
opt: Options = Options(false)

scala> opt.iLikeNulls
res2: Boolean = false
```

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -249,6 +256,17 @@ case class NumMatchesAndCount(numMatches: Long, count: Long, override val fullCo
...
+}
+
+object NullBehavior extends Enumeration {
+  type NullBehavior = Value
+  val Ignore, Empty, Fail = Value
```
rdsharma26, line 267: What does `Empty` mean for numbers? If it is only applicable for Strings, should we call it `EmptyString` ?

src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
```diff
@@ -42,5 +48,19 @@ case class MaxLength(column: String, where: Option[String] = None)
...
+    nullBehavior match {
+      case NullBehavior.Fail =>
+        val colLengths: Column = length(conditionalSelection(column, where)).cast(DoubleType)
+        conditionalSelection(colLengths, Option(s"${column} IS NULL"), replaceWith = Double.MaxValue)
+      case NullBehavior.Empty =>
+        length(conditionalSelection(col(column), Option(s"${column} IS NULL"), replaceWith = "")).cast(DoubleType)
+      case _ => length(conditionalSelection(column, where)).cast(DoubleType)
+    }
+  }
+  private def getNullBehavior(): NullBehavior = {
```
rdsharma26, line 61: nit: The `()` can be removed. In idiomatic Scala, `()`, called an empty paren(theses) method, denotes that function performs some side effect. Like calling a service or even just logging. Here, we are just extracting data from an object, and therefore, we can remove this `()` from the method.

src/main/scala/com/amazon/deequ/analyzers/MinLength.scala
```diff
@@ -41,4 +48,22 @@ case class MinLength(column: String, where: Option[String] = None)
   }
 
   override def filterCondition: Option[String] = where
+
+  @VisibleForTesting
```
rdsharma26, line 52: Is this needed if we are using package private modifier and have the test in the appropriate package?

mentekid (approved):

src/test/scala/com/amazon/deequ/VerificationSuiteTest.scala
```diff
@@ -199,11 +201,79 @@ class VerificationSuiteTest extends WordSpec with Matchers with SparkContextSpec
...
+      assert(result.status == CheckStatus.Error)
+
+      val resultData = VerificationResult.rowLevelResultsAsDataFrame(session, result, data)
+
+      resultData.show()
+      val expectedColumns: Set[String] =
+        data.columns.toSet + expectedColumn1 + expectedColumn2
+      assert(resultData.columns.toSet == expectedColumns)
+
+
```
rdsharma26, line 236: nit: Excess whitespace

rdsharma26 (approved):
LGTM. Left some minor comments.

PR #466, opened 2023-04-12
New API added to referential integrity to allow for row level annotation

rdsharma26:
*Issue #, if available:*
- No associated issue. This is a WIP experimental utility.

*Description of changes:*
- The provided primary data frame will have a column added to it to indicate a true/false value.
- The value will be true if the values of the provided columns for that row exist in the reference dataframe. Otherwise, the value will be false.
- Added various tests, including tests for nested columns.
- Refactored the parameter validation from the regular row level check, so that the same validation can be done in the new function as well.
- Updated scaladoc of the two APIs.

### DIFF
--- a/src/test/scala/com/amazon/deequ/comparison/ReferentialIntegrityTest.scala
+++ b/src/test/scala/com/amazon/deequ/comparison/ReferentialIntegrityTest.scala
@@ -45,11 +45,10 @@ class ReferentialIntegrityTest extends AnyWordSpec with SparkContextSpec {
         (6, "Megan", "TX")))
       val ds2 = rdd2.toDF("new_id", "name", "state")
 
-      val col1 = "name_foo"
-      val col2 = "name"
+      val col = "name"
       val assertion: Double => Boolean = _ >= 1.0
 
-      val result = ReferentialIntegrity.subsetCheck(ds1, Seq.empty, ds2, Seq(col2), assertion)
+      val result = ReferentialIntegrity.subsetCheck(ds1, Seq.empty, ds2, Seq(col), assertion)
       assert(result.isInstanceOf[ComparisonFailed])
       assert(result.asInstanceOf[ComparisonFailed].errorMessage.contains(s"Empty list provided"))
     }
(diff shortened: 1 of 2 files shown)


src/test/scala/com/amazon/deequ/comparison/ReferentialIntegrityTest.scala
```diff
@@ -474,4 +558,123 @@ class ReferentialIntegrityTest extends AnyWordSpec with SparkContextSpec {
...
+
+      val rdd2 = spark.sparkContext.parallelize(Seq(
+        (1, "New York", "NY"),
+        (2, "Wisconsin", "WI"),
+        (3, "Texas", "TX"),
+        (4, "California", "CA"))) // Reference has correct row
+      val ds2 = rdd2.toDF("id", "state name", "state")
+
+      val cols = Seq("state name", "state")
+      val outcomeCol = "row_level_outcome"
```
mentekid, line 581: Can you add an assertion for the columns returned? It should be exactly `("id", "state name", "state", "row_level_outcome")`

src/test/scala/com/amazon/deequ/comparison/ReferentialIntegrityTest.scala
```diff
@@ -474,4 +558,123 @@ class ReferentialIntegrityTest extends AnyWordSpec with SparkContextSpec {
...
+
+      val result = ReferentialIntegrity.subsetCheckRowLevel(ds1, cols, ds2, cols, Some(outcomeCol))
+      assert(result.isRight)
+      val outcomes = result.right.get.select(outcomeCol).collect().toSeq.map { r => r.get(0) }
+      assert(outcomes == Seq(true, true, false, true))
```
eycho-am, line 586: I'm a little confused as to why it's `true, true, false, true` when the last row is the incorrect one

mentekid, line 586: maybe also assert that if I drop `"row_level_outcome"` I get a dataframe with exactly what's in `rdd1`?

rdsharma26: Fixed.

src/test/scala/com/amazon/deequ/comparison/ReferentialIntegrityTest.scala
rdsharma26: Done.

src/test/scala/com/amazon/deequ/comparison/ReferentialIntegrityTest.scala
rdsharma26: Done.

rdsharma26:
> Would it be possible to also add tests in the `VerificationSuiteTests` to see how the row level results for ref integrity are added with other row level results?

This is a standalone utility and we are not using it in the VerificationSuite. Once the verification suite supports multiple dataframes, we will add this check and integrate the row level results.

eycho-am (approved):
LGTM

PR #468, opened 2023-04-20
Feature: ColumnValues to Row-level Results

zixianzh1:
*Issue #, if available:*

*Description of changes:*
- Added Min, Max, PatternMatch to Rowlevel results.
- Added nullBehavior to Min and Max, specific behavior may needs further discussion.
- Added Compliance to Rowlevel results.

TODO:
- Confirm the behavior of Compliance

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -460,7 +460,7 @@ private[deequ] object Analyzers {
   }
 
   def entityFrom(columns: Seq[String]): Entity.Value = {
-    if (columns.size == 1) Entity.Column else Entity.Mutlicolumn
+    if (columns.size == 1) Entity.Column else Entity.Multicolumn
   }
 
   def conditionalSelection(selection: String, where: Option[String]): Column = {
--- a/src/main/scala/com/amazon/deequ/analyzers/Correlation.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Correlation.scala
@@ -67,7 +67,7 @@ case class Correlation(
     secondColumn: String,
     where: Option[String] = None)
   extends StandardScanShareableAnalyzer[CorrelationState]("Correlation",
-    s"$firstColumn,$secondColumn", Entity.Mutlicolumn)
+    s"$firstColumn,$secondColumn", Entity.Multicolumn)
   with FilterableAnalyzer {
 
   override def aggregationFunctions(): Seq[Column] = {
--- a/src/main/scala/com/amazon/deequ/analyzers/MutualInformation.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/MutualInformation.scala
@@ -16,12 +16,14 @@
 
 package com.amazon.deequ.analyzers
 
+import com.amazon.deequ.analyzers.Analyzers.COUNT_COL
 import com.amazon.deequ.analyzers.Analyzers._
-import com.amazon.deequ.metrics.{DoubleMetric, Entity}
-import org.apache.spark.sql.functions.{col, sum, udf}
+import com.amazon.deequ.metrics.DoubleMetric
+import com.amazon.deequ.metrics.Entity
+import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.functions.sum
+import org.apache.spark.sql.functions.udf
 import org.apache.spark.sql.types.StructType
-import Analyzers.COUNT_COL
-import com.amazon.deequ.analyzers.runners.MetricCalculationException
 
 /**
   * Mutual Information describes how much information about one column can be inferred from another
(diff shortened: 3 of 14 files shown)


src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -336,7 +336,13 @@ object Constraint {
       case _ => s"PatternMatchConstraint($column, $pattern)"
     }
 
-    new NamedConstraint(constraint, constraintName)
+    val sparkAssertion = org.apache.spark.sql.functions.udf(assertion)
+//    new NamedConstraint(constraint, constraintName)
```
zixianzh1, line 340: will remove this

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -260,7 +260,7 @@ case class AnalyzerOptions(nullBehavior: NullBehavior = NullBehavior.Ignore)
 
 object NullBehavior extends Enumeration {
   type NullBehavior = Value
-  val Ignore, EmptyString, Fail = Value
+  val Ignore, EmptyString, Fail, NegVal = Value
```
mentekid, line 263: It would probably be better to separate this into two enums, one for strings and a different for columns. That's because `EmptyString` makes no sense for numbers and `NegVal` makes no sense for strings. As it stands now, if I pass `NullBehavior.EmptyString` to the `Maximum` analyzer, I get no warning and unexpected behavior.

Another option would be to implement a method inside the enum that is given a Spark type and returns the sentinel value for that type (if one is defined) or throws. For example, for [numeric types](https://spark.apache.org/docs/3.2.2/sql-ref-datatypes.html), it returns `-1`. For String type, it returns `""`. For anything else, it throws. I am not entirely sure that would work though, as I don't know if Analyzers know the type they're working on

There may also be a way to let users override the default sentinel but may not be worth implementing in this PR if it's too complicated

mentekid: On another note, I think `Ignore` and `Fail` are misleading as names (I know you didn't create this enum but as I look at it this occurs to me).

What we really mean is, `Ignore` will produce a result that doesn't take into account `null` values, and `Fail` will preserve `null` values as they are and that may fail down the line. May be worth renaming `Fail` to `Preserve`, as an opposite to "Replace" in this context

src/main/scala/com/amazon/deequ/analyzers/Minimum.scala
```diff
@@ -55,4 +57,22 @@ case class Minimum(column: String, where: Option[String] = None)
   }
 
   override def filterCondition: Option[String] = where
+
+  @VisibleForTesting
+  private def criterion(nullBehavior: NullBehavior): Column = {
```
mentekid, line 62: Since we're duplicating code here, consider moving this to a trait such as `NullReplacingAnalyzer` or `ExtremumAnalyzer` that is implemented by `Minimum` and `Maximum` (Extremum is the general term for a max or min)

src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
```diff
@@ -77,6 +74,14 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] =
     }
   }
 
+  @VisibleForTesting // required by some tests that compare analyzer results to an expected state
+  private[deequ] def criterion(): Column = {
```
mentekid, line 78: nit: You don't need `()` when functions take no arguments in Scala

src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
```diff
@@ -77,6 +74,14 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] =
...
+    val expression = when(regexp_extract(col(column), pattern.toString(), 0) =!= lit(""), 1)
+      .otherwise(0)
+    conditionalSelection(expression, where).cast(IntegerType)
```
mentekid, line 81: this could be a boolean, right? We don't need to assert it further

src/test/scala/com/amazon/deequ/analyzers/PatternMatchTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    "return row-level results for columns" in withSparkSession { session =>
+
+      val data = getDfWithStringColumns(session)
+
+      val patternMatchCountry = PatternMatch("Address Line 1", """\d""".r)
+      val state = patternMatchCountry.computeStateFrom(data)
+      val metric: DoubleMetric with FullColumn = patternMatchCountry.computeMetricFrom(state)
+
+      data.withColumn("new", metric.fullColumn.get).collect().map(_.getAs[Integer]("new")) shouldBe
+        Seq(1, 1, 1, 1, 1, 1, 1, 1)
```
mentekid, line 37: I'd expect this to be a boolean as noted above

Also, can you have a few lines where this returns `false` (or if it needs to be an integer, returns other values e.g. 0 and 2)

mentekid (commented):
This is great. I left some comments around null behavior now that we have seen this is a problem in both numbers and strings.

Issue #381, opened 2021-09-07
Enhancement Request: Histogram with aggregate on a second column


rpasupat:
We have a requirement to store an aggregate (sum for example) of a "second" column along with the histogram counts. Would like to know if such a requirement is already being considered. 

I want to analyse the trends based on the second columns aggregate (while picking the top N entries from the histogram). 

I imagine the implementation being the addition of a new metric case class and a new analyzer which does a second column aggregate (in addition to count) and filling the metric (as part of compute). 

Suggestions on using current features to achieve this is much appreciated.

twollnik:
Just to clarify: You want to track an aggregate metric for each of the historam bins, is that right? So, this would be logically similar to e.g. `data.groupBy("firstColumn").agg(count("*"), sum("secondColumn"))`.

I am not sure how this could be achieved with the current API. @lange-labs @tdhd @TammoR Can you weigh in?

rpasupat:
Yes exactly. Thanks very much @twollnik 

Eagerly look forward to a solution from deequ experts here. Either a solution/workaround using existing APIs or this being accepted as a good requirement (but can't achieve using existing features). It will help me confirm and make a decision on my requirement.

rpasupat:
@twollnik Thanks for your response. Appreciate it. 

If I understand it right, you are suggesting to store details on the aggregate (e.g. Sum) separately and then compare it with the histogram bins. Is that right?

We store a set of histogram metrics in the metrics repository and later for new data compare it what is in history. Doing it custom would mean, storing similar entries in a different file/database for sum and then compare it with what we retrieve from metrics repo. If we go the custom route (direct spark route I mean), I would rather store both "count" and "sum" side by side making my job a lot easier. 

I will read through the other mentioned tickets here.

PR #475, opened 2023-04-29
Alternative aggregate functions to calculate histogram values.

akalotkin:
*Issue #, if available:*
NA. Kind of close to #381 

*Description of changes:*
- Histogram analyzer changes
  -  Provide a way to calculate histogram values using different aggregate functions (sum).
  - Add aggregateColumn - a column castable to Long. Aggregate function is applied to this column
- Histogram serialization changes
  - Add aggregateFunction property ("count" is default in case of missing property)
  - Add aggregateColumn property

In addition to
``` 
select(col(column).cast(StringType)).groupBy(column).count()
```
provide a way get histogram values
```
select(col(column).cast(StringType), col(aggregateColumn.get).cast(LongType)).groupBy(column).sum(aggregateColumn.get)
```

WIP. Please take a look at the changes. I'll add more tests, docs and validation logic if you think that these changes make sense.

Thanks.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
@@ -138,15 +138,6 @@ object Histogram {
   val count_function = "count"
   val sum_function = "sum"
 
-
-  def createAggregateFunction(function: String, aggregateColumn: String): AggregateFunction = {
-    function match {
-      case Histogram.count_function => Count
-      case Histogram.sum_function => Sum(aggregateColumn)
-      case _ => throw new IllegalArgumentException("Wrong aggregate function name: " + function)
-    }
-  }
-
   sealed trait AggregateFunction {
     def query(column: String, data: DataFrame): DataFrame
 
@@ -196,40 +187,6 @@ object Histogram {
 
     override def function(): String = sum_function
   }
-
-//  case class Min(aggColumn: String) extends AggregateFunction {
-//    override def query(column: String, data: DataFrame): DataFrame = {
-//      data
-//        .select(col(column).cast(StringType), col(aggColumn).cast(LongType))
-//        .na.fill(Histogram.NullFieldReplacement)
-//        .groupBy(column)
-//        .min(aggColumn)
-//        .withColumnRenamed("count", Analyzers.COUNT_COL)
-//    }
-//
-//    override def aggregateColumn(): Option[String] = {
-//      Some(aggColumn)
-//    }
-//
-//    override def function(): String = min
-//  }
-//
-//  case class Max(aggColumn: String) extends AggregateFunction {
-//    override def query(column: String, data: DataFrame): DataFrame = {
-//      data
-//        .select(col(column).cast(StringType), col(aggColumn).cast(LongType))
-//        .na.fill(Histogram.NullFieldReplacement)
-//        .groupBy(column)
-//        .max(aggColumn)
-//        .withColumnRenamed("count", Analyzers.COUNT_COL)
-//    }
-//
-//    override def aggregateColumn(): Option[String] = {
-//      Some(aggColumn)
-//    }
-//
-//    override def function(): String = max
-//  }
 }
 
 
--- a/src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
@@ -189,9 +189,6 @@ object Histogram {
   }
 }
 
-
-
-
 object OrderByAbsoluteCount extends Ordering[Row] {
   override def compare(x: Row, y: Row): Int = {
     x.getLong(1).compareTo(y.getLong(1))
--- a/src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
@@ -20,8 +20,9 @@ import com.amazon.deequ.analyzers.runners.{IllegalAnalyzerParameterException, Me
 import com.amazon.deequ.metrics.{Distribution, DistributionValue, HistogramMetric}
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions.col
-import org.apache.spark.sql.types.{StringType, StructType}
+import org.apache.spark.sql.types.{DoubleType, LongType, StringType, StructType}
 import org.apache.spark.sql.{DataFrame, Row}
+
 import scala.util.{Failure, Try}
 
 /**
@@ -43,7 +44,9 @@ case class Histogram(
     binningUdf: Option[UserDefinedFunction] = None,
     maxDetailBins: Integer = Histogram.MaximumAllowedDetailBins,
     where: Option[String] = None,
-    computeFrequenciesAsRatio: Boolean = true)
+    computeFrequenciesAsRatio: Boolean = true,
+    aggregateFunction: String = Histogram.Count,
+    aggregateColumn: Option[String] = None)
   extends Analyzer[FrequenciesAndNumRows, HistogramMetric]
   with FilterableAnalyzer {
 
@@ -63,14 +66,10 @@ case class Histogram(
       1
     }
 
-    val frequencies = data
+    val df = data
       .transform(filterOptional(where))
       .transform(binOptional(binningUdf))
-      .select(col(column).cast(StringType))
-      .na.fill(Histogram.NullFieldReplacement)
-      .groupBy(column)
-      .count()
-      .withColumnRenamed("count", Analyzers.COUNT_COL)
+    val frequencies = query(df)
 
     Some(FrequenciesAndNumRows(frequencies, totalCount))
   }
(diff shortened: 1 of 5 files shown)


akalotkin:
@rdsharma26, can you please take a look? Thanks.

rdsharma26:
@akalotkin Will be having a look shortly. In the meantime, could you fix the build failure?

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -43,7 +44,9 @@ case class Histogram(
     binningUdf: Option[UserDefinedFunction] = None,
     maxDetailBins: Integer = Histogram.MaximumAllowedDetailBins,
     where: Option[String] = None,
-    computeFrequenciesAsRatio: Boolean = true)
+    computeFrequenciesAsRatio: Boolean = true,
+    aggregateFunction: String = Histogram.Count,
```
rdsharma26, line 48: Can we make this an enum? In Scala 2, we can use a `sealed trait` and `case object` to create enumerations.

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -43,7 +44,9 @@ case class Histogram(
...
+    aggregateColumn: Option[String] = None)
```
rdsharma26, line 49: This is used when the aggregate function is sum. As a caller of this API, it is possible that this value is not provided when using `sum`. Is it possible to group this configuration into a single object? 

Check this parameter for reference: https://github.com/awslabs/deequ/commit/ef4c308a165d5928dc61e2349d907a232f8761f9#diff-4db9bbf900d2a35e16c36a158cbf3c4641d336e1930e6f343f2fef5773d0cea6R95

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -125,11 +124,33 @@ case class Histogram(
       case _ => data
     }
   }
+
+  private def query(data: DataFrame): DataFrame = {
+    aggregateFunction match {
+      case Histogram.Count => countQuery(data)
+      case Histogram.Sum => sumQuery(data)
+      case _ => throw new IllegalStateException()
```
rdsharma26, line 132: If we use an enumeration, then we won't need a default case.

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -125,11 +124,33 @@ case class Histogram(
...
+    }
+  }
+
+  private def countQuery(data: DataFrame): DataFrame = {
+      data.select(col(column).cast(StringType))
+      .na.fill(Histogram.NullFieldReplacement)
+      .groupBy(column).count().withColumnRenamed("count", Analyzers.COUNT_COL)
```
rdsharma26, line 139: Can we use the same formatting as the one on lines 66-73 originally?

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -125,11 +124,33 @@ case class Histogram(
...
+  }
+
+  private def sumQuery(data: DataFrame): DataFrame = {
+    data.select(col(column).cast(StringType), col(aggregateColumn.get).cast(LongType))
+      .na.fill(Histogram.NullFieldReplacement)
+      .groupBy(column).sum(aggregateColumn.get).withColumnRenamed("count", Analyzers.COUNT_COL)
```
rdsharma26, line 145: Can we use the same formatting as the one on lines 66-73 originally?

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -125,13 +126,115 @@ case class Histogram(
...
+  val sum_function = "sum"
+
+
+  def createAggregateFunction(function: String, aggregateColumn: String): AggregateFunction = {
+    function match {
+      case Histogram.count_function => Count
+      case Histogram.sum_function => Sum(aggregateColumn)
+      case _ => throw new IllegalArgumentException("Wrong aggregate function name: " + function)
+    }
+  }
```
rdsharma26, line 148: Can we remove this function and let the caller of the function provide a valid instance of `AggregateFunction` ? The reason I say this is because if we add more subclasses of `AggregateFunction` in the future, we will need to update this `createAggregateFunction` to accept the superset of ALL parameters for all subtypes.

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
```diff
@@ -125,13 +126,115 @@ case class Histogram(
...
+//        .max(aggColumn)
+//        .withColumnRenamed("count", Analyzers.COUNT_COL)
+//    }
+//
+//    override def aggregateColumn(): Option[String] = {
+//      Some(aggColumn)
+//    }
+//
+//    override def function(): String = max
+//  }
```
rdsharma26, line 232: Either we remove the commented out code, or we keep it as code and add tests. Otherwise, it adds bloat to the source code.

src/main/scala/com/amazon/deequ/repository/AnalysisResultSerde.scala
```diff
@@ -302,6 +302,10 @@ private[deequ] object AnalyzerSerializer
         result.addProperty(ANALYZER_NAME_FIELD, "Histogram")
         result.addProperty(COLUMN_FIELD, histogram.column)
         result.addProperty("maxDetailBins", histogram.maxDetailBins)
+        if (histogram.aggregateFunction != Histogram.Count) {
```
rdsharma26, line 305: Can we add a comment here that:
- the reason we are excluding `Count` since it is the current default
- It was the only previously supported function in the past, and we exclude it for backwards compatibility

src/test/scala/com/amazon/deequ/analyzers/runners/AnalyzerContextTest.scala
```diff
@@ -86,13 +86,13 @@ class AnalyzerContextTest extends AnyWordSpec
               |{"entity":"Column","instance":"item","name":"Distinctness","value":1.0},
               |{"entity":"Column","instance":"att1","name":"Completeness","value":1.0},
               |{"entity":"Multicolumn","instance":"att1,att2","name":"Uniqueness","value":0.25},
+              |{"entity":"Dataset","instance":"*","name":"Size (where: att2 == 'd')","value":1.0},
+              |{"entity":"Dataset","instance":"*","name":"Size","value":4.0},
```
rdsharma26, line 90: Is this change required? It is not clear from the changes that the order of the fields in the serialization process is changing. If it is not, can we remove this diff to make the PR clearer ?

akalotkin: Yes.  This is required. This is the reason a test was failing.

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
akalotkin: I guess. I just forgot to remove this code. Will do

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
akalotkin: This function is used to deserialize object from json. I can move it to AnalysisResultSerde, but there must be a piece of code that deserializes all combinations from json.

src/main/scala/com/amazon/deequ/repository/AnalysisResultSerde.scala
akalotkin: Will add comments.

src/main/scala/com/amazon/deequ/analyzers/Histogram.scala
rdsharma26: Sure, I think it is fine then to keep it outside of `Histogram` and in `AnalysisResultSerde`.

src/test/scala/com/amazon/deequ/analyzers/runners/AnalyzerContextTest.scala
rdsharma26: Thanks for the information. Is it possible to understand why it is required? The reason I ask is because of backwards compatibility. Will a serialized object that has the older order fail to get deserialized because of this change?

akalotkin: I need to dig deeper to explain why it happens. My guess is that there is some kind of HashMap, HashSet that changes order of response.

akalotkin: Yes. That's what happens. Below are screenshots from 2 different versions (breakpoint is on this line https://github.com/awslabs/deequ/blob/f53283e05ca93a16f054d4783e000082699fe0c3/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala#L201)

As you can see after map union it gives two different results. And code here (https://github.com/awslabs/deequ/blob/f53283e05ca93a16f054d4783e000082699fe0c3/src/main/scala/com/amazon/deequ/analyzers/runners/AnalyzerContext.scala#L83) iterates through the HashMap. 


Master branch
master
feature/histogram-aggregate branch
feature:histogram-aggregate

akalotkin:
@rdsharma26, all done.

rdsharma26:
Thanks for the PR! @akalotkin

rdsharma26 (approved):

PR #493, opened 2023-07-05
Replace Spark SQL isNull check with Spark Scala based DSL

rdsharma26:
*Issue*
- The column profiling for min/max lengths for a string column is not working when the column name has a space in it.

*Description of changes:*
- Replaced Spark SQL isNull check with Spark Scala based DSL
- This is to ensure columns with spaces in their names get their names escaped correctly in the where condition.
- Added a test to verify.


eycho-am (approved):
lgtm

PR #508, opened 2023-10-04
Allow all DQ constraints to be generated from an Analyzer

mentekid:
*Issue #, if available:*

*Description of changes:*

Due to our typing system, it is always possible to create an Analyzer, but converting that into a generic Constraint and generic Check is impossible. 

Instead, give callers the option to convert an Analyzer into a Check with these helper functions. This enables reuse of Analyzers - you can use an Analyzer directly, or convert it to a Check by assigning it a severity and assertion.


eycho-am (approved):
lgtm

rdsharma26 (approved):

PR #510, opened 2023-10-11
Fail when CustomSql has syntax errors

mentekid:
*Issue #, if available:*

*Description of changes:*

If the SQL statement has an error, return a failed state from the CustomSql analyzer


mentekid:
I'm going to relax the test expectation because it will start failing on different spark versions.

PR #512, opened 2023-10-24
Creation of Exact Quantile Check

jmilis2000:
*Issue #, if available:*

*Description of changes:*
Added hasExactQuantile Check. Similar to hasApproxQuantile but finds more precise quantile value.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/ExactQuantile.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/ExactQuantile.scala
@@ -1,0 +1,49 @@
+package com.amazon.deequ.analyzers
+
+import com.amazon.deequ.analyzers.Preconditions.{hasColumn, isNumeric}
+import com.amazon.deequ.analyzers.Analyzers.{conditionalSelection, ifNoNullsIn}
+import com.amazon.deequ.metrics.FullColumn
+import org.apache.curator.shaded.com.google.common.annotations.VisibleForTesting
+import org.apache.spark.sql.{Column, Row}
+import org.apache.spark.sql.functions.expr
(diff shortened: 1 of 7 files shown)


rdsharma26:
@jmilis2000 thanks for the PR! We will be having a look shortly. Can you please check in the meantime why the build is failing?

jmilis2000:
> @jmilis2000 thanks for the PR! We will be having a look shortly. Can you please check in the meantime why the build is failing?

@rdsharma26 I fixed the issue. The build should work now. It failed because there was no amazon copyright at the top of the new analyzer i created.

rdsharma26 (approved):
LGTM. To confirm, you are using Spark's `percentile` to calculate the exact percentile rather than using the existing `hasApproxQuantile` check with error = `0.0`. I assume this is because the latter uses Spark's approx quantile method.

src/main/scala/com/amazon/deequ/analyzers/ExactQuantile.scala
```diff
@@ -0,0 +1,65 @@
...
+ * permissions and limitations under the License.
+ *
+ */
+
+package com.amazon.deequ.analyzers
+
+import com.amazon.deequ.analyzers.Preconditions.{hasColumn, isNumeric}
+import com.amazon.deequ.analyzers.Analyzers.{conditionalSelection, ifNoNullsIn}
+import com.amazon.deequ.metrics.FullColumn
+import org.apache.curator.shaded.com.google.common.annotations.VisibleForTesting
```
rdsharma26, line 22: We use `import com.google.common.annotations.VisibleForTesting` in other places.

PR #515, opened 2023-10-26
[Data Synchronization/Matching] Delegate to Spark for checking existence of columns in the given dataframes

rdsharma26:
*Description of changes:*

- Prior to this change, we were doing case sensitive equality checks of non-key columns.
- This makes the utility more restrictive, as Spark does not care about the casing of column names.
- With this change, we rely on Spark to check if a column exists in the given dataframe. If Spark can find the column, we can proceed with the rest of the check.

*Issue #, if available:*
N/A


eycho-am (approved):
LGTM

Issue #502, opened 2023-08-21
[FEATURE] Add spark table metric repository


charlieyou:
**Is your feature request related to a problem? Please describe.**
Given that the rest of Deequ relies on Spark, it seems incongruous that there is no support for loading metrics from a Spark table. Saving to a JSON works fine for now, but as we scale up, we would like to take advantage of the data catalog/governance that comes along with using Spark tables (specifically with Databricks in our case, but can imagine it being generally useful outside of that).

**Describe the solution you'd like**
An implementation of `MetricsRepository` using Spark tables as the data source.

**Describe alternatives you've considered**
This can be hacked together by dumping a spark table to a JSON file and then reading that with the FS MR, but it's quite inelegant.

**Additional context**
Happy to take a crack at the implementation myself when I have more capacity in a few days.

VenkataKarthikP:
I can take a stab at this, just want to check if PR's are accepted of this issue? @rdsharma26

VenkataKarthikP:
@mentekid @charlieyou  i worked on #518  to implement this request, please take a look. thanks in advance.

VenkataKarthikP:
@mentekid @rdsharma26 can we close this as we have PR merged. Also, can we get a release tag with latest changes. Thanks.

sudsali:
Closing as implemented. `SparkMetricsRepository` exists at `src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala` and allows reading/writing metrics via Spark tables.

PR #518, opened 2023-10-30
MetricsRepository using Spark tables as the data source

VenkataKarthikP:
*Issue #, if available:*
https://github.com/awslabs/deequ/issues/502 
*Description of changes:*
An implementation of MetricsRepository using Spark tables as the data source.

Currently, DQ heavily relies on spark for processing. But to save metrics to a table like Iceberg or Hive or other spark catalog based, we don't have an implementation. With this PR, users can  save metrics to a table and read them offline as needed. 

Example usage - 

```scala
val spark = <your spark session>
val repository = new SparkTableMetricsRepository(spark, "metrics_table")

repository.save(resultKey, context)

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,6 +155,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.iceberg</groupId>
+            <artifactId>iceberg-spark-runtime-3.3_2.12</artifactId>
+            <version>0.14.0</version>
+            <scope>test</scope>
+        </dependency>
 
     </dependencies>
 
--- a/src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
+++ b/src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
@@ -1,0 +1,135 @@
+/**
+ * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ * use this file except in compliance with the License. A copy of the License
+ * is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
+ */
+package com.amazon.deequ.repository.sparktable
+
+import com.amazon.deequ.analyzers.Analyzer
+import com.amazon.deequ.analyzers.runners.AnalyzerContext
+import com.amazon.deequ.metrics.Metric
+import com.amazon.deequ.repository._
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Row, SaveMode, SparkSession}
+
+class SparkTableMetricsRepository(session: SparkSession, tableName: String) extends MetricsRepository {
+
+  private val SCHEMA = StructType(Array(
+    StructField("result_key", StringType),
+    StructField("metric_name", StringType),
+    StructField("metric_value", StringType),
+    StructField("result_timestamp", StringType),
+    StructField("serialized_context", StringType)
+  ))
+
+  override def save(resultKey: ResultKey, analyzerContext: AnalyzerContext): Unit = {
+    val serializedContext = AnalysisResultSerde.serialize(Seq(AnalysisResult(resultKey, analyzerContext)))
+
+    val rows = analyzerContext.metricMap.map { case (analyzer, metric) =>
+      Row(resultKey.toString, analyzer.toString, metric.value.toString,
+        resultKey.dataSetDate.toString, serializedContext)
+    }.toSeq
+
+    val metricDF = session.createDataFrame(session.sparkContext.parallelize(rows), SCHEMA)
+
+    metricDF.write
+      .mode(SaveMode.Append)
+      .saveAsTable(tableName)
+  }
+
+  override def loadByKey(resultKey: ResultKey): Option[AnalyzerContext] = {
+    val df: DataFrame = session.table(tableName)
+    val matchingRows = df.filter(col("result_key") === resultKey.toString).collect()
+
+    if (matchingRows.isEmpty) {
+      None
+    } else {
+      val serializedContext = matchingRows(0).getAs[String]("serialized_context")
+      val analysisResult = AnalysisResultSerde.deserialize(serializedContext).head
+      Some(analysisResult.analyzerContext)
+    }
+  }
+
+  override def load(): MetricsRepositoryMultipleResultsLoader = {
+    SparkTableMetricsRepositoryMultipleResultsLoader(session, tableName)
+  }
+
+}
+
+
+case class SparkTableMetricsRepositoryMultipleResultsLoader(session: SparkSession,
+                                                            tableName: String,
+                                                            tagValues: Option[Map[String, String]] = None,
+                                                            analyzers: Option[Seq[Analyzer[_, Metric[_]]]] = None,
+                                                            timeAfter: Option[Long] = None,
+                                                            timeBefore: Option[Long] = None
+                                                           ) extends MetricsRepositoryMultipleResultsLoader {
+
+  override def withTagValues(tagValues: Map[String, String]): MetricsRepositoryMultipleResultsLoader =
+    this.copy(tagValues = Some(tagValues))
+
+  override def forAnalyzers(analyzers: Seq[Analyzer[_, Metric[_]]]): MetricsRepositoryMultipleResultsLoader =
+    this.copy(analyzers = Some(analyzers))
+
+  override def after(dateTime: Long): MetricsRepositoryMultipleResultsLoader =
+    this.copy(timeAfter = Some(dateTime))
+
+  override def before(dateTime: Long): MetricsRepositoryMultipleResultsLoader =
+    this.copy(timeBefore = Some(dateTime))
+
+  override def get(): Seq[AnalysisResult] = {
+    val initialDF: DataFrame = session.table(tableName)
+
+    initialDF.printSchema()
+    val tagValuesFilter: DataFrame => DataFrame = df => {
+      tagValues.map { tags =>
+        tags.foldLeft(df) { (currentDF, tag) =>
+          currentDF.filter(row => {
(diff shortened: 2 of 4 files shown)


rdsharma26:
Hi @VenkataKarthikP, thank you for the PR! While we review the PR, could you please provide more details on the use case that this PR helps enable? An example snippet of code of how this feature can be used will be great, to both serve as motivation for the PR and for documentation going forward.

VenkataKarthikP:
> Hi @VenkataKarthikP, thank you for the PR! While we review the PR, could you please provide more details on the use case that this PR helps enable? An example snippet of code of how this feature can be used will be great, to both serve as motivation for the PR and for documentation going forward.

@rdsharma26 updated PR description with more details, let me know if it looks good.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
...
+    StructField("serialized_context", StringType)
+  ))
+
+  override def save(resultKey: ResultKey, analyzerContext: AnalyzerContext): Unit = {
+    val serializedContext = AnalysisResultSerde.serialize(Seq(AnalysisResult(resultKey, analyzerContext)))
+
+    val rows = analyzerContext.metricMap.map { case (analyzer, metric) =>
+      Row(resultKey.toString, analyzer.toString, metric.value.toString,
+        resultKey.dataSetDate.toString, serializedContext)
+    }.toSeq
```
rdsharma26, line 42: Do need to check for successful metrics before converting to rows?

The following is being done in the other repositories:

```
val successfulMetrics = analyzerContext.metricMap.filter {
  case (_, metric) => metric.value.isSuccess
}
```

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
...
+
+    val metricDF = session.createDataFrame(session.sparkContext.parallelize(rows), SCHEMA)
```
rdsharma26, line 44: We can create a case class for the schema and use `toDF` in order to avoid having to create a sequence of rows and a DF using that.

```
case class SparkTableMetric(resultKey: String, metricName: String, metricValue: String)

import session.implicits._

val metricsDF = Seq(
  SparkTableMetric("key_1", "Completeness", 1.0),
  SparkTableMetric("key_2", "Uniqueness", 0.5),
).toDF()

```

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  override def loadByKey(resultKey: ResultKey): Option[AnalyzerContext] = {
+    val df: DataFrame = session.table(tableName)
+    val matchingRows = df.filter(col("result_key") === resultKey.toString).collect()
+
+    if (matchingRows.isEmpty) {
+      None
+    } else {
+      val serializedContext = matchingRows(0).getAs[String]("serialized_context")
+      val analysisResult = AnalysisResultSerde.deserialize(serializedContext).head
```
rdsharma26, line 59: We should `.headOption` to be safer.

```
AnalysisResultSerde.deserialize(serializedContext).headOption.map(_.analyzerContext)
```

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
...
+
+}
+
+
+case class SparkTableMetricsRepositoryMultipleResultsLoader(session: SparkSession,
+                                                            tableName: String,
+                                                            tagValues: Option[Map[String, String]] = None,
+                                                            analyzers: Option[Seq[Analyzer[_, Metric[_]]]] = None,
+                                                            timeAfter: Option[Long] = None,
+                                                            timeBefore: Option[Long] = None
```
rdsharma26, line 76: For consistency purposes, can we use the pattern as the other repositories? That includes changing these to private fields.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
...
+    this.copy(tagValues = Some(tagValues))
+
+  override def forAnalyzers(analyzers: Seq[Analyzer[_, Metric[_]]]): MetricsRepositoryMultipleResultsLoader =
+    this.copy(analyzers = Some(analyzers))
+
+  override def after(dateTime: Long): MetricsRepositoryMultipleResultsLoader =
+    this.copy(timeAfter = Some(dateTime))
+
+  override def before(dateTime: Long): MetricsRepositoryMultipleResultsLoader =
+    this.copy(timeBefore = Some(dateTime))
```
rdsharma26, line 89: We can update these to be consistent with the other repositories.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  override def get(): Seq[AnalysisResult] = {
+    val initialDF: DataFrame = session.table(tableName)
+
+    initialDF.printSchema()
```
rdsharma26, line 94: Can we remove this?

src/test/scala/com/amazon/deequ/SparkContextSpec.scala
```diff
@@ -80,11 +99,22 @@ trait SparkContextSpec {
       .config("spark.sql.shuffle.partitions", 2.toString)
       .config("spark.sql.adaptive.enabled", value = false)
       .config("spark.driver.bindAddress", "127.0.0.1")
+      .config("spark.sql.warehouse.dir", warehouseDir.toAbsolutePath.toString)
```
rdsharma26, line 102: Should this be added conditionally, where Iceberg support is required? This change affects all spark sessions used by all unit tests.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,135 @@
+/**
+ * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
rdsharma26, line 2: nit: Can you fix the year in all new files added in this PR?

src/test/scala/com/amazon/deequ/repository/sparktable/SparkTableMetricsRepositoryTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      val repository = new SparkTableMetricsRepository(spark, "metrics_table")
+      // Save the metric
+      repository.save(resultKey, context)
+
+      // Load the metric
+      val loadedContext = repository.loadByKey(resultKey)
+
+      assert(loadedContext.isDefined)
+      assert(loadedContext.get.metric(analyzer).contains(metric))
+    }
```
rdsharma26, line 52: Please remove all such braces, since they are not required.

pom.xml
```diff
@@ -155,6 +155,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.iceberg</groupId>
+            <artifactId>iceberg-spark-runtime-3.3_2.12</artifactId>
+            <version>0.14.0</version>
+            <scope>test</scope>
+        </dependency>
```
rdsharma26, line 164: Correct me if I'm wrong, but we don't seem to be using any specific feature of Iceberg. If that's the case, can we remove this dependency ?

VenkataKarthikP: in tests I have added, test to check if Iceberg table writes are working as expected. So that's the reason to add this dep in test scope.

src/test/scala/com/amazon/deequ/SparkContextSpec.scala
VenkataKarthikP: It's not just for Iceberg, its used for other spark tables writes as well. I have updated to add this only when needed. 

If we do not set this when we run local tests, spark warehouse dir will get created in project root directory when executing unit tests.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
VenkataKarthikP: My intent was to avoid creating, `var` variables and mutate them. Instead wanted to utilize copy method provided by case classes to create a new object. I can update this implementation in other places as well in follow up PR.

If you strongly suggest sticking with implementation in other repositories, let me know.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
VenkataKarthikP: same as bellow + based on below I can make these `private val` to avoid accessing outside.

VenkataKarthikP:
@rdsharma26 thanks for review, have addressed review comments, with some follow up questions.

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
```diff
@@ -0,0 +1,130 @@
...
+      .filter { case (_, metric) => metric.value.isSuccess }
+
+    val metricDF = successfulMetrics.map { case (analyzer, metric) =>
+      SparkTableMetric(resultKey.toString, analyzer.toString, metric.value.toString,
+        resultKey.dataSetDate, serializedContext)
+    }.toSeq.toDF()
+
+    metricDF.write
+      .mode(SaveMode.Append)
+      .saveAsTable(tableName)
```
SirWerto, line 42: Hello @VenkataKarthikP, thank you so much for the added feature!!!

Could we consider to let the user define the write options in some way?

VenkataKarthikP: sure, I think it's great idea. let me check!

src/main/scala/com/amazon/deequ/repository/sparktable/SparkMetricsRepository.scala
rdsharma26: @VenkataKarthikP That is fair. I agree with that approach as well. The concern I have is with the consistency across the repositories. We can address that in a later PR.

rdsharma26 (approved):

VenkataKarthikP:
@rdsharma26 can we merge this PR. thanks

Issue #521, opened 2023-11-13
[FEATURE] Exposing Anomaly Strategy Calculation Thresholds for Users


arsenalgunnershubert777:
**Is your feature request related to a problem? Please describe.**
Right now Anomaly Checks only return a success or fail response, but I'd like to retrieve the thresholds used in the Anomaly Strategy calculations. This would help DeeQu users clearly see what numbers the Anomaly Check actually used for the calculation.

**Describe the solution you'd like**
I'd like to have the Constraint Result return a field that contains thresholds used in the Anomaly Strategy. I'm also planning to make a pull request to implement this feature, and would love to hear your feedback on that.

**Describe alternatives you've considered**
N/A

**Additional context**
My current plan for implementing this feature is as follows:
1. [isNewestPointNonAnomalous function](https://github.com/awslabs/deequ/blob/54c5e48c21ebbe30a75c142e4621d5ca4669ab48/src/main/scala/com/amazon/deequ/checks/Check.scala#L429) currently is an assertion function that takes in a Double metric and returns a Boolean of if there are anomalies detected. Change this function to return an AnomalyAssertionResult which contains a Boolean and also Doubles to represent thresholds used in the Anomaly Strategy calculations.
2. Create an AnomalyBasedConstraint where the [pickValueAndAssert function](https://github.com/awslabs/deequ/blob/54c5e48c21ebbe30a75c142e4621d5ca4669ab48/src/main/scala/com/amazon/deequ/constraints/AnalysisBasedConstraint.scala#L67) will retrieve the Boolean from the anomalyAssertionResult.
3. Create an AnomalyConstraintResult with a field (or fields) for those thresholds, where the pickValueAndAssert function passes in those thresholds from the anomalyAssertionResult. Now the user can view the results.
4. In Constraint, make the [anomalyConstraint function](https://github.com/awslabs/deequ/blob/54c5e48c21ebbe30a75c142e4621d5ca4669ab48/src/main/scala/com/amazon/deequ/constraints/Constraint.scala#L227) use the AnomalyBasedConstraint class.
5. Make any necessary upstream or downstream changes to make the functionality work.
6. User inheritance, traits, and try to make the code as clean as possible.

PR #525, opened 2023-12-13
Exposing Helpful Anomaly Detection Metadata from Anomaly Strategies (ie Anomaly Thresholds)

arsenalgunnershubert777:
*Issue #, if available:*
[521](https://github.com/awslabs/deequ/issues/521)

*Description of changes:*

This PR adds functionality to expose anomaly detection metadata for anomaly checks

- AnomalyDetectionMetadata is a newly exposed field that contains information about the newest point that was run through the Anomaly Check, what the data value is, what the value being checked by the Anomaly Strategy is, what the thresholds used in the Anomaly Check are, if the data point was an Anomaly. The user is able to see these details from ConstraintResult.
- There are the following changes to make this happen
  - First of all the Anomaly class was changed into [AnomalyDetectionDataPoint class](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L34)
    - This class added an anomalyMetricValue (Double), anomalyThreshold (AnomalyThreshold), isAnomaly (boolean)
      - [anomalyThreshold](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L28) is single [AnomalyThreshold](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L93) object
        - The AnomalyThreshold class has two [Bound](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L101C1-L101C1) classes, for upper and lower bounds 
        - The Bound class contains a Double value and boolean if the Bound is inclusive or not
      - [anomalyMetricValue](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L24) is the Double value being compared to the anomalyThresholds, this most of the time is the same as the data value but not always (like in rate of change strategies where the anomalyMetricValue is the change between data points)
      - [isAnomaly](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L30) boolean is added because now the Anomaly Strategies output not just Anomaly data points, but all data points with Anomaly Check related information, this is so that the AnomalyThresholds can get passed to the user whether or not the data point was found to be an actual Anomaly. For example, see the [BatchNormalStrategy](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/BatchNormalStrategy.scala#L84-L94)
      - value (Double) field was renamed to [dataMetricValue](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L23) to distinguish from anomalyMetricValue and indicate this is the data point value from the data series, it was also changed from Option[Double] because the Anomaly Check used a [Vector of Doubles ](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/AnomalyDetectionStrategy.scala#L30)
      - I added some comments, but I’m not too familiar about the confidence field (I see mostly 1.0 filled in for that), so I left a [TODO](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L31) to fill in that definition in the class comments
  - As mentioned previously, [all the Anomaly Strategies](https://github.com/arsenalgunnershubert777/deequ/tree/anomaly-threshold-expose/src/main/scala/com/amazon/deequ/anomalydetection) now output all the AnomalyDetectionDataPoints with those extra Anomaly details corresponding to all the data points in the search interval, the isAnomaly boolean identifies if that data point is an anomaly
  - These AnomalyDetectionDataPoints are passed along the [AnomalyDetectionResult](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L110) (renamed from DetectionResult) back through the [Anomaly assertion function](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/checks/Check.scala#L438). 
    - This assertion function uses a new [AnomalyBasedConstraint class](https://github.com/arsenalgunnershubert777/deequ/blob/anomaly-threshold-expose/src/main/scala/com/amazon/deequ/constraints/AnomalyBasedConstraint.scala) to handle the assertion function [returning both an assertion boolean as well as the AnomalyDetectionDataPoint](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/constraints/AnomalyBasedConstraint.scala#L50) of the newest data point. 
  - Since the anomaly check only checks the newest data point, there is returned a wrapper class called [AnomalyDetectionMetadata](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/anomalydetection/DetectionResult.scala#L134) that contains that AnomalyDetectionDataPoint.
    - The above functionality is in a new [getNewestPointAnomalyResults](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/checks/Check.scala#L1193) function in Check
    - In the future if the Anomaly Check focuses on multiple datapoints instead of just one, the AnomalyDetectionMetadata can be adjusted accordingly.
  - The anomaly detection metadata is [added to the ConstraintResult class](https://github.com/arsenalgunnershubert777/deequ/blob/847d522eaba94e0e276240ef7631c15e85d2a149/src/main/scala/com/amazon/deequ/constraints/Constraint.scala#L47) as an optional field
    - Maybe this can be separated into a different class that inherits from a common trait since the AnomalyDetectionMetadata does not pertain to other non anomaly checks
  - Tests are updated to reflect the above changes

Would love to hear any thoughts, feedback, things to change. Whether relating to overall design, or small renaming and code changes. Thanks so much!


rdsharma26:
Thank you @arsenalgunnershubert777 for the PR and happy new year! We will be reviewing the PR this week.

arsenalgunnershubert777:
@rdsharma26 thanks for the heads up! I am working on fixing the merge conflicts, should be done soon

rdsharma26:
@arsenalgunnershubert777 Thanks again for the PR and for your patience while we reviewed this PR.

Having looked deeper into the changes, I have one high level comment. This PR introduces changes that are backwards incompatible. Any consumer of Deequ who is using the `Anomaly` class or the `isNewestPointNonAnomalous` method in `Check.scala` will see their code fail to compile when they upgrade to this version of Deequ.

Could you update the PR such that the new class is added alongsde `Anomaly`? We can also create a new method called `isNewestPointNonAnomalousWithExtendedResult` which uses your new class as its return type.

arsenalgunnershubert777:
Hi @rdsharma26 thank you for the response. I have some questions for clarity. I will try to make changes to create new class and new method and work for backwards compatibility. But I'm wondering, in the VerificationRunBuilder, in order to make the Anomaly Check use the new feature, I would have to also update this [getAnomalyCheck method](https://github.com/awslabs/deequ/blob/f3222c1f0077ef3729a703117b391211d5e645e9/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L310) to use .isNewestPointAnomalousWithExtendedResult right? Would that make affect backwards compatibility? Let me know if my understanding of any of this is correct, thanks again!

rdsharma26:
> Hi @rdsharma26 thank you for the response. I have some questions for clarity. I will try to make changes to create new class and new method and work for backwards compatibility. But I'm wondering, in the VerificationRunBuilder, in order to make the Anomaly Check use the new feature, I would have to also update this [getAnomalyCheck method](https://github.com/awslabs/deequ/blob/f3222c1f0077ef3729a703117b391211d5e645e9/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L310) to use .isNewestPointAnomalousWithExtendedResult right? Would that change affect backwards compatibility as well? Let me know if my understanding of any of this is correct, thanks again!

Instead of updating, you will add a new method, which then uses the new `isNewestPointAnomalousWithExtendedResult` method. Think of the following 3 scenarios:

1. Anyone who is using Deequ's anomaly detection faces no interruption when upgrading their Deequ version to the one that contains your changes.
2. Anyone who wishes to use Deequ's anomaly detection can use your new method and reference the documentation on which one of the two methods to pick (vanilla results vs extended results). Therefore, do update the documentation as well.
3. Anyone who is using Deequ's anomaly detection feature and wishes to use your new feature can simply switch to the new method and everything should continue to work as expected. Again, updated documentation will help users with the transition.

arsenalgunnershubert777:
@rdsharma26 got it, that makes sense! I think I will add an optional parameter in the [getAnomalyCheckMethod](https://github.com/awslabs/deequ/blob/f3222c1f0077ef3729a703117b391211d5e645e9/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L302) to choose which method to use, and it can default to original method. That way everything should be compatible

rdsharma26:
> @rdsharma26 got it, that makes sense! I think I will add an optional parameter in both [addAnomalyCheck](https://github.com/awslabs/deequ/blob/f3222c1f0077ef3729a703117b391211d5e645e9/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L227) and [getAnomalyCheckMethod](https://github.com/awslabs/deequ/blob/f3222c1f0077ef3729a703117b391211d5e645e9/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala#L302) to choose which method to use, and it can default to original method. That way everything should be compatible

Sounds good. Looks like there is `AnomalyCheckConfig` available where this new config parameter could be added.

arsenalgunnershubert777:
Hi @rdsharma26 hope you are well, I created [new PR to make this backwards compatible](https://github.com/awslabs/deequ/pull/593) thanks!

sudsali:
Closing this PR since #593 is the PR with backward compatible function as requested.

Issue #501, opened 2023-08-18
Incorporate referential integrity and data synchronization checks into Deequ's VerificationSuite


rdsharma26:
The following two utilities should be part of Deequ's verification suite.

- https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/comparison/DataSynchronization.scala
- https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/comparison/ReferentialIntegrity.scala

VenkataKarthikP:
@rdsharma26 could you add more details on what enhancement we are looking at here, I could take a stab at implementation. thanks

mentekid:
Hi,

Here's some context for this issue. Deequ lets you run checks on your data by constructing a [VerificationSuite](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/VerificationSuite.scala) object. You can build a `VerificationSuite` by calling `onData()`, which then exposes the `addCheck()` method. You can call that repeatedly to add multiple checks to your suite, for example:


```
val verificationResult = VerificationSuite()
  .onData(data)
  .addCheck(Check(CheckLevel.Error, "must have 5 rows").hasSize(_ == 5)
  .addCheck(Check(CheckLevel.Error, "must have no nulls").isComplete("id")
  .run()
```

All checks within the same Verification suite are processed before Spark is called, and Deequ comes up with a plan to calculate all the necessary statistics without making unnecessary passes over the data.

However, for the comparison operations @rdsharma26 is mentioning above, there is no `Check` object and therefore they cannot be added to a `VerificationSuite`. This means two things:

1. There is no way for Deequ to optimize the execution of these comparisons
2. The code to run any of these checks looks incongruent with any other Deequ syntax, as you need to directly invoke methods rather than constructing a cohesive suite of tests.

The ask here is to merge the two operations with the standard Deequ APIs, so a user can create a verification suite that contains a mix of cross-dataset and in-dataset tests. This will probably require a bit of refactoring in the [VerificationRunBuilder](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/VerificationRunBuilder.scala), because unlike any `Check`s we have today, the cross-dataset checks require an additional reference dataset in addition to the primary dataset (passed using `onData()`, which returns a `VerificationRunBuilder`).

Let me know if that's not clear or you have any follow-up questions.

VenkataKarthikP:
[...]

PR #526, opened 2023-12-26
add data synchronization test to verification Suite.

VenkataKarthikP:
*Issue, if available: #501

*Description of changes:* Adding data synchronization check to verification suite, with this change users can define `isDataSynchronized` check.

Example usage - 
```scala

val verificationResult = VerificationSuite()
  .onData(data)
  .addCheck(Check(CheckLevel.Error, "must have data in sync")
                                      .isDataSynchronized(dfToCompare, Map("id" -> "id"), _ > 0.7)
  .run()
```

cc: @mentekid @rdsharma26

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -19,21 +19,15 @@ package com.amazon.deequ.analyzers
 import com.amazon.deequ.analyzers.Analyzers._
 import com.amazon.deequ.analyzers.NullBehavior.NullBehavior
 import com.amazon.deequ.analyzers.runners._
-import com.amazon.deequ.metrics.FullColumn
+import com.amazon.deequ.comparison.{DataSynchronization, DataSynchronizationFailed, DataSynchronizationSucceeded}
+import com.amazon.deequ.metrics.{DoubleMetric, Entity, FullColumn, Metric}
 import com.amazon.deequ.utilities.ColumnUtil.removeEscapeColumn
-import com.amazon.deequ.metrics.DoubleMetric
-import com.amazon.deequ.metrics.Entity
-import com.amazon.deequ.metrics.Metric
+import org.apache.spark.sql.{Column, DataFrame, Row, SparkSession}
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types._
-import org.apache.spark.sql.Column
-import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.Row
-import org.apache.spark.sql.SparkSession
 
 import scala.language.existentials
-import scala.util.Failure
-import scala.util.Success
+import scala.util.{Failure, Success, Try}
 
 /**
   * A state (sufficient statistic) computed from data, from which we can compute a metric.
@@ -298,6 +292,56 @@ abstract class GroupingAnalyzer[S <: State[_], +M <: Metric[_]] extends Analyzer
     groupingColumns().map { name => Preconditions.hasColumn(name) } ++ super.preconditions
   }
 }
+
+/**
+ * Data Synchronization Analyzer
+ *
+ * @param dfToCompare    DataFrame to compare
+ * @param columnMappings columns mappings
+ * @param assertion      assertion logic
+ */
+case class DataSynchronizationAnalyzer(dfToCompare: DataFrame,
+                                       columnMappings: Map[String, String],
+                                       assertion: Double => Boolean)
+  extends Analyzer[DataSynchronizationState, DoubleMetric] {
+
+  override def computeStateFrom(data: DataFrame): Option[DataSynchronizationState] = {
+
+    val result = DataSynchronization.columnMatch(data, dfToCompare, columnMappings, assertion)
+
+    result match {
+      case succeeded: DataSynchronizationSucceeded =>
+        Some(DataSynchronizationState(succeeded.passedCount.getOrElse(0), succeeded.totalCount.getOrElse(0)))
+      case failed: DataSynchronizationFailed =>
+        Some(DataSynchronizationState(failed.passedCount.getOrElse(0), failed.totalCount.getOrElse(0)))
+      case _ => None
+    }
+  }
+
+  override def computeMetricFrom(state: Option[DataSynchronizationState]): DoubleMetric = {
+
+    state match {
+      case Some(s) => DoubleMetric(
+        Entity.Dataset,
+        "DataSynchronization",
+        "",
+        Try(s.synchronizedDataCount.toDouble / s.dataCount.toDouble),
+        None
+      )
+      case None => DoubleMetric(
+        Entity.Dataset,
+        "DataSynchronization",
+        "",
+        Try(0.0),
+        None
+      )
+    }
+  }
+
+  override private[deequ] def toFailureMetric(failure: Exception) =
+    metricFromFailure(failure, "DataSynchronization", "", Entity.Dataset)
+}
+
 
 /** Helper method to check conditions on the schema of the data */
 object Preconditions {
--- a/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may not
  * use this file except in compliance with the License. A copy of the License
@@ -19,15 +19,21 @@ package com.amazon.deequ.analyzers
 import com.amazon.deequ.analyzers.Analyzers._
 import com.amazon.deequ.analyzers.NullBehavior.NullBehavior
 import com.amazon.deequ.analyzers.runners._
-import com.amazon.deequ.comparison.{DataSynchronization, DataSynchronizationFailed, DataSynchronizationSucceeded}
-import com.amazon.deequ.metrics.{DoubleMetric, Entity, FullColumn, Metric}
+import com.amazon.deequ.metrics.DoubleMetric
+import com.amazon.deequ.metrics.Entity
+import com.amazon.deequ.metrics.FullColumn
+import com.amazon.deequ.metrics.Metric
 import com.amazon.deequ.utilities.ColumnUtil.removeEscapeColumn
-import org.apache.spark.sql.{Column, DataFrame, Row, SparkSession}
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types._
 
 import scala.language.existentials
-import scala.util.{Failure, Success, Try}
+import scala.util.Failure
+import scala.util.Success
 
 /**
   * A state (sufficient statistic) computed from data, from which we can compute a metric.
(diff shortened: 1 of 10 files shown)


src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -19,21 +19,15 @@ package com.amazon.deequ.analyzers
...
 import org.apache.spark.sql.types._
-import org.apache.spark.sql.Column
-import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.Row
-import org.apache.spark.sql.SparkSession
 
 import scala.language.existentials
-import scala.util.Failure
-import scala.util.Success
+import scala.util.{Failure, Success, Try}
```
rdsharma26, line 30: We avoid block imports as they can lead to merge conflicts. Each import on its own line makes that process easier. Can we remove this diff?

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -299,6 +293,56 @@ abstract class GroupingAnalyzer[S <: State[_], +M <: Metric[_]] extends Analyzer
...
 }
 
+/**
+ * Data Synchronization Analyzer
+ *
+ * @param dfToCompare    DataFrame to compare
+ * @param columnMappings columns mappings
+ * @param assertion      assertion logic
+ */
+case class DataSynchronizationAnalyzer(dfToCompare: DataFrame,
```
rdsharma26, line 303: This should be in its own file like the other analyzers.

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -299,6 +293,56 @@ abstract class GroupingAnalyzer[S <: State[_], +M <: Metric[_]] extends Analyzer
   }
 }
 
+/**
+ * Data Synchronization Analyzer
+ *
+ * @param dfToCompare    DataFrame to compare
```
rdsharma26, line 299: nit: `DataFrame to compare with`

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -299,6 +293,56 @@ abstract class GroupingAnalyzer[S <: State[_], +M <: Metric[_]] extends Analyzer
...
+ * @param columnMappings columns mappings
```
rdsharma26, line 300: nit: We can add more documentation here. Are these the key column mappings? What about the other parameter, comparison column mappings? Will that be added in a future PR?

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -299,6 +293,56 @@ abstract class GroupingAnalyzer[S <: State[_], +M <: Metric[_]] extends Analyzer
...
+        Try(s.synchronizedDataCount.toDouble / s.dataCount.toDouble),
+        None
+      )
+      case None => DoubleMetric(
+        Entity.Dataset,
+        "DataSynchronization",
+        "",
+        Try(0.0),
+        None
+      )
```
rdsharma26, line 337: We can simplify it as follows: 

```
val metric = state match {
  case Some(s) => Try(s.synchronizedDataCount.toDouble / s.dataCount.toDouble).getOrElse(0.0)
  case _ => 0.0
}

DoubleMetric(Entity.Dataset, "DataSynchronization", "", metric, None)
```

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -27,6 +26,7 @@ import com.amazon.deequ.repository.MetricsRepository
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import com.amazon.deequ.anomalydetection.HistoryUtils
 import com.amazon.deequ.checks.ColumnCondition.{isAnyNotNull, isEachNotNull}
+import org.apache.spark.sql.DataFrame
```
rdsharma26, line 29: Same comment as one above. Please avoid block imports.

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -338,6 +338,38 @@ case class Check(
...
+   *                        of the synchronization check. This object can be used in Deequ's verification suite to
+   *                        assert data quality constraints.
+   */
+  def isDataSynchronized(otherDf: DataFrame, columnMappings: Map[String, String], assertion: Double => Boolean,
+                         hint: Option[String] = None): Check = {
+
+    val dataSyncAnalyzer = DataSynchronizationAnalyzer(otherDf, columnMappings, assertion)
+    val constraint = DataSynchronizationConstraint(dataSyncAnalyzer, hint)
+    addConstraint(constraint)
+
```
rdsharma26, line 370: nit: excess whitespace

src/main/scala/com/amazon/deequ/comparison/ComparisonResult.scala
```diff
@@ -17,5 +17,11 @@
...
-case class ComparisonFailed(errorMessage: String) extends ComparisonResult
-case class ComparisonSucceeded() extends ComparisonResult
+
+case class ComparisonFailed(errorMessage: String, ratio: Double = 0) extends ComparisonResult
+case class ComparisonSucceeded(ratio: Double = 0) extends ComparisonResult
+
+case class DataSynchronizationFailed(errorMessage: String, passedCount: Option[Long] = None,
+                                     totalCount: Option[Long] = None) extends ComparisonResult
+case class DataSynchronizationSucceeded(passedCount: Option[Long] = None, totalCount: Option[Long] = None)
+  extends ComparisonResult
```
rdsharma26, line 27: If the check succeeded, do we expect the `passedCount` and `totalCount` to be `None`? If not, can we keep their types as `Long`?

src/main/scala/com/amazon/deequ/comparison/ComparisonResult.scala
```diff
@@ -17,5 +17,11 @@
 package com.amazon.deequ.comparison
 
 sealed trait ComparisonResult
-case class ComparisonFailed(errorMessage: String) extends ComparisonResult
-case class ComparisonSucceeded() extends ComparisonResult
+
+case class ComparisonFailed(errorMessage: String, ratio: Double = 0) extends ComparisonResult
+case class ComparisonSucceeded(ratio: Double = 0) extends ComparisonResult
```
rdsharma26, line 22: We should keep 1 set of states. Right now, there are 2 sets of states. `Comparison[Failed/Succeeded]` vs `DataSynchronization[Failed/Succeeded]`. Having too many states can result in confusion for the end user.

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -1092,7 +1124,10 @@ case class Check(
         case nc: ConstraintDecorator => nc.inner
         case c: Constraint => c
       }
-      .collect { case constraint: AnalysisBasedConstraint[_, _, _] => constraint.analyzer }
+      .collect {
+        case constraint: AnalysisBasedConstraint[_, _, _] => constraint.analyzer
+        case constraint: DataSynchronizationConstraint => constraint.analyzer
```
rdsharma26, line 1129: Why can't the existing statement match `DataSynchronizationConstraint` ?

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -17,9 +17,10 @@
 package com.amazon.deequ.constraints
 
 import com.amazon.deequ.analyzers._
-import com.amazon.deequ.metrics.{BucketDistribution, Distribution, Metric}
+import com.amazon.deequ.metrics.{BucketDistribution, Distribution, DoubleMetric, Metric}
 import org.apache.spark.sql.expressions.UserDefinedFunction
 
+import scala.util.{Success, Try}
```
rdsharma26, line 23: Please avoid block imports.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -897,3 +898,29 @@ object Constraint {
...
+ * Data Synchronization Constraint
+ * @param analyzer Data Synchronization Analyzer
+ * @param hint hint
+ */
+case class DataSynchronizationConstraint(analyzer: DataSynchronizationAnalyzer, hint: Option[String])
+  extends Constraint {
+
+  override def evaluate(metrics: Map[Analyzer[_, Metric[_]], Metric[_]]): ConstraintResult = {
+
+    val anz = Try(metrics.filter(i => i._1.isInstanceOf[DataSynchronizationAnalyzer]).head._2)
```
rdsharma26, line 912: Can we write it as the following, for better readability?

```
val (_, anz) = Try(metrics.filter { case(analyzer, _)  => analyzer.isInstanceOf[DataSynchronizationAnalyzer] }.head)
```

What happens if `.head` is called on empty list? Is that possible?

rdsharma26 (commented):
Thank you @VenkataKarthikP for this PR! I've left some comments.

Note: We use this check for the `DatasetMatch` rule in AWS Glue Data Quality. In a future PR, we can rename `DataSynchronization` to `DatasetMatch` for consistency purposes.

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
VenkataKarthikP: thanks, will update documentation. 
I was planning to do this in phased manner instead of one big PR, will definitely do a follow up PR to add that as well.

src/main/scala/com/amazon/deequ/comparison/ComparisonResult.scala
VenkataKarthikP: yes, changed to `Long` type.

src/main/scala/com/amazon/deequ/comparison/ComparisonResult.scala
VenkataKarthikP: I was planning, to consolidate once `ReferentialIntegrity` is also integrated into verification suite. I will handle that in next PR.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
VenkataKarthikP: will refactor to make it improve readability, thanks for suggestion.

src/main/scala/com/amazon/deequ/checks/Check.scala
VenkataKarthikP: updated, yes we don't need this.

VenkataKarthikP:
@rdsharma26 thanks for the review, updated with review comments.

src/main/scala/com/amazon/deequ/analyzers/DataSynchronizationAnalyzer.scala
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * An Analyzer for Deequ that performs a data synchronization check between two DataFrames.
+ * It evaluates the degree of synchronization based on specified column mappings and an assertion function.
+ *
+ * The analyzer computes a ratio of synchronized data points to the total data points, represented as a DoubleMetric.
+ * Refer to [[com.amazon.deequ.comparison.DataSynchronization.columnMatch]] for DataSynchronization implementation
+ *
+ * @param dfToCompare The DataFrame to compare with the primary DataFrame that is setup
+ *                    during [[com.amazon.deequ.VerificationSuite.onData]] setup.
+ * @param columnMappings A map where each key-value pair represents a column in the primary DataFrame
```
rdsharma26, line 40: Please rename to `keyColumnMappings`.

src/main/scala/com/amazon/deequ/analyzers/DataSynchronizationAnalyzer.scala
```diff
@@ -0,0 +1,94 @@
...
+ *                                                                              _ > 0.8).run()
+ *
+ *
+ * The computeStateFrom method calculates the synchronization state by comparing the specified columns of the two
+ * DataFrames.
+ * The computeMetricFrom method then converts this state into a DoubleMetric representing the synchronization ratio.
+ *
+ */
+case class DataSynchronizationAnalyzer(dfToCompare: DataFrame,
+                                       columnMappings: Map[String, String],
```
rdsharma26, line 64: Please rename to `keyColumnMappings`.

rdsharma26 (approved):
LGTM. Thanks again for the PR! Please address the following in next PR.

- `columnMappings` should be renamed to `keyColumnMappings`
- Add `matchColumnMappings`
- [If possible] Change `DataSynchronization` to `DatasetMatch`

Issue #530, opened 2024-01-16
[FEATURE] Filter condition is ignored when filtering records based on row-level checks


eapframework:
I tried the filtering data based on row-level checks and it is working fine. New column is added with true or false based on the row-level checks. But if rules has filter condition that has been ignored.

eycho-am:
Hi @eapframework, thank you for raising this issue. Could you provide an example of the behavior you're seeing and the behavior you expect/want instead?

We're actually looking into a similar issue - here's an example unit test using the `Completeness` analyzer:

For the dataframe:
```
  def getDfCompleteAndInCompleteColumns(sparkSession: SparkSession): DataFrame = {
    import sparkSession.implicits._

    Seq(
      ("1", "a", "f"),
      ("2", "b", "d"),
      ("3", "a", null),
      ("4", "a", "f"),
      ("5", "b", null),
      ("6", "a", "f")
    ).toDF("item", "att1", "att2")
  }
```
This test checks the row-level results
```
    "return row-level results for columns filtered" in withSparkSession { session =>

      val data = getDfCompleteAndInCompleteColumns(session)

      val completenessAtt2 = Completeness("att2", Option("att1 = \"a\""))
      val state = completenessAtt2.computeStateFrom(data)
      val metric: DoubleMetric with FullColumn = completenessAtt2.computeMetricFrom(state)

      data.withColumn("new", metric.fullColumn.get).collect().map(_.getAs[Boolean]("new")) shouldBe
        Seq(true, false, false, true, false, true)
    }
```
Using the verification suite on a similar test:
```
+----+----+----+-----+
|item|att1|att2|rule1|
+----+----+----+-----+
|   1|   a|   f| true|
|   2|   b|   d|false|
|   3|   a|null|false|
|   4|   a|   f| true|
|   5|   b|null|false|
|   6|   a|   f| true|
+----+----+----+-----+
```

Here we can see that the rows that EITHER are filtered out (rows 2,5 att1 is not `a`) or fail the check (row 3 is null) are marked as false.
Would you expect rows 2,5 to show true/None in this case?

eapframework:
Thanks for your response. This is the same issue I am facing.  
I am expecting rows 2,5 to show true because those are not failed records

Expected result:
+----+----+----+-----+
|item|att1|att2|rule1|
+----+----+----+-----+
|   1|   a|   f| true|
|   2|   b|   d|true|
|   3|   a|null|false|
|   4|   a|   f| true|
[...]

PR #532, opened 2024-02-13
Feature: Add Row Level Result Treatment Options for Uniqueness and Completeness

eycho-am:
*Issue #, if available:* #530

*Description of changes:*
This PR adds the trait `RowLevelFilterTreatement` that enables users to define how filtered rows will be labeled as when retrieving row level results. The two options are `True` and `Null` and this defaults to Null.

The `RowLevelFilterTreatement` can be defined in the `VerificationRunBuilder` like below:
```
val suite = new VerificationSuite().onData(data)
  .withRowLevelFilterTreatment(FilteredRow.NULL)
  .addChecks(<your_checks>)
```

This PR defines the behavior for the `Completeness` and `Uniqueness` analyzers, and will be updated for other analyzers in future PRs.


src/main/scala/com/amazon/deequ/utilities/RowLevelFilterTreatement.scala
```diff
@@ -0,0 +1,48 @@
...
+trait RowLevelFilterTreatment {
+  def rowLevelFilterTreatment: FilteredRow
+}
+
+/**
+ * Companion object for RowLevelFilterTreatment
+ * Defines a sharedInstance that can be used throughout the VerificationRunBuilder
+ */
+object RowLevelFilterTreatment {
+  private var _sharedInstance: RowLevelFilterTreatment = new RowLevelFilterTreatmentImpl(FilteredRow.TRUE)
```
mentekid, line 32: if VerificationRunSuite has-a RowLevelFilterTreatment, then this variable could be absorbed into the class, which already contains several mutable members. That keeps all mutation code in the same place - this is how we add analyzers for example:

```Scala
protected var requiredAnalyzers: Seq[Analyzer[_, Metric[_]]] = Seq.empty 

...

def addRequiredAnalyzers(requiredAnalyzers: Seq[Analyzer[_, Metric[_]]]): this.type = {
  this.requiredAnalyzers ++= requiredAnalyzers
  this
}

```

mentekid (approved):

src/main/scala/com/amazon/deequ/analyzers/Analyzer.scala
```diff
@@ -62,7 +67,7 @@ trait DoubleValuedState[S <: DoubleValuedState[S]] extends State[S] {
 }
 
 /** Common trait for all analyzers which generates metrics from states computed on data frames */
-trait Analyzer[S <: State[_], +M <: Metric[_]] extends Serializable {
+trait Analyzer[S <: State[_], +M <: Metric[_]] extends Serializable with RowLevelFilterTreatment {
```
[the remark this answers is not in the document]
eycho-am: Removed `RowLevelFilterTreatement` and using `AnalyzerOptions` instead

src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
```diff
@@ -88,7 +90,15 @@ object FrequencyBasedAnalyzer {
       .count()
 
     // Set rows with value count 1 to true, and otherwise false
-    val fullColumn: Column = count(UNIQUENESS_ID).over(Window.partitionBy(columnsToGroupBy: _*))
+    val fullColumn: Column = {
+      val window = Window.partitionBy(columnsToGroupBy: _*)
+      where.map {
+        condition => {
+          count(when(expr(condition), UNIQUENESS_ID)).over(window)
+        }
```
rdsharma26, line 98: nit: we can remove the brackets after `=>`

src/main/scala/com/amazon/deequ/analyzers/UniqueValueRatio.scala
```diff
@@ -34,11 +38,27 @@ case class UniqueValueRatio(columns: Seq[String], where: Option[String] = None)
...
+    toSuccessMetric(numUniqueValues / numDistinctValues, fullColumnUniqueness)
   }
 
   override def filterCondition: Option[String] = where
+
+  private def getRowLevelFilterTreatment: FilteredRow = {
+    analyzerOptions
+      .map { options => options.filteredRow }
+      .getOrElse(FilteredRow.TRUE)
+  }
```
rdsharma26, line 61: nit: This is repeated in a few places, so could go into the base class.

src/main/scala/com/amazon/deequ/analyzers/UniqueValueRatio.scala
```diff
@@ -34,11 +38,27 @@ case class UniqueValueRatio(columns: Seq[String], where: Option[String] = None)
...
     val numUniqueValues = result.getDouble(offset)
     val numDistinctValues = result.getLong(offset + 1).toDouble
-    val fullColumnUniqueness = when((fullColumn.getOrElse(null)).equalTo(1), true).otherwise(false)
-    toSuccessMetric(numUniqueValues / numDistinctValues, Option(fullColumnUniqueness))
+    val conditionColumn = where.map { expression => expr(expression) }
+    val fullColumnUniqueness = fullColumn.map {
+      rowLevelColumn => {
+        conditionColumn.map {
+          condition => {
+            when(not(condition), expr(getRowLevelFilterTreatment.toString))
```
rdsharma26, line 46: Same comment for `expr(getRowLevelFilterTreatment.toString)` as above

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -314,6 +340,22 @@ case class Check(
...
+   *                        Refers to the fraction of unique values.
+   * @param hint            A hint to provide additional context why a constraint could have failed
+   * @param analyzerOptions Options to configure analyzer behavior (NullTreatment, FilteredRow)
+   * @return
+   */
+  def hasUniqueness(column: String, assertion: Double => Boolean, hint: Option[String],
+                    analyzerOptions: Option[AnalyzerOptions])
+  : CheckWithLastConstraintFilterable = {
+    hasUniqueness(Seq(column), assertion, hint, analyzerOptions)
+  }
```
rdsharma26, line 357: Can't we combine all the `hasUniqueness` into one method and update the body of the method based on the existence of parameters?

src/test/scala/com/amazon/deequ/analyzers/runners/AnalyzerContextTest.scala
```diff
@@ -145,7 +145,8 @@ class AnalyzerContextTest extends AnyWordSpec
   }
 
   private[this] def assertSameJson(jsonA: String, jsonB: String): Unit = {
-    assert(SimpleResultSerde.deserialize(jsonA) ==
-      SimpleResultSerde.deserialize(jsonB))
+    assert(SimpleResultSerde.deserialize(jsonA).toSet.sameElements(SimpleResultSerde.deserialize(jsonB).toSet))
+    //    assert(SimpleResultSerde.deserialize(jsonA) ==
+    //      SimpleResultSerde.deserialize(jsonB))
```
rdsharma26, line 150: Can we remove these comments here and in other files below?

src/test/scala/com/amazon/deequ/analyzers/runners/AnalysisRunnerTests.scala
```diff
@@ -195,7 +200,14 @@ class AnalysisRunnerTests extends AnyWordSpec
...
-        assert(separateResults.toString == runnerResults.toString)
+        // assert(separateResults == runnerResults.toString)
+        // Used to be tested with the above line, but adding filters changed the order of the results.
+        assert(separateResults.asInstanceOf[Set[DoubleMetric]].size ==
+          runnerResults.asInstanceOf[Set[DoubleMetric]].size)
+        separateResults.asInstanceOf[Set[DoubleMetric]].foreach( result => {
+            assert(runnerResults.toString.contains(result.toString))
+          }
+          )
       }
```
rdsharma26, line 211: This block of code has some excess whitespace and misaligned brackets. In a future PR, can we add these rules into scala-style such that the build fails otherwise?

PR #535, opened 2024-02-20
Feature/row level filter min max

eycho-am:
*Issue #, if available:* N/A

*Description of changes:*
- Follow-up to PR #532 
- Adding AnalyzerOptions to use FileteredRowOutcome for Minimum and Maximum analyzers

### DIFF
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -681,15 +681,17 @@ case class Check(
     * @param column Column to run the assertion on
     * @param assertion Function that receives a double input parameter and returns a boolean
     * @param hint A hint to provide additional context why a constraint could have failed
+    * @param analyzerOptions Options to configure analyzer behavior (NullTreatment, FilteredRow)
     * @return
     */
   def hasMin(
       column: String,
       assertion: Double => Boolean,
-      hint: Option[String] = None)
+      hint: Option[String] = None,
+      analyzerOptions: Option[AnalyzerOptions] = None)
     : CheckWithLastConstraintFilterable = {
 
-    addFilterableConstraint { filter => minConstraint(column, assertion, filter, hint) }
+    addFilterableConstraint { filter => minConstraint(column, assertion, filter, hint, analyzerOptions) }
   }
 
   /**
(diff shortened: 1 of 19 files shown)


src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
```diff
@@ -48,15 +51,35 @@ case class MaxLength(column: String, where: Option[String] = None, analyzerOptio
...
+    val whereNotCondition = where.map { expression => not(expr(expression)) }
+    rowOutcome match {
+      case FilteredRowOutcome.TRUE =>
+        conditionSelectionGivenColumn(col, whereNotCondition, replaceWith = Double.MinValue)
+      case _ =>
+        conditionSelectionGivenColumn(col, whereNotCondition, replaceWith = null)
+    }
+  }
+
+  private def transformColForNullBehavior(col: Column, nullBehavior: NullBehavior): Column = {
```
rdsharma26, line 73: Can we use `column` here directly instead of passing it in as`col` ? In `rowLevelResults`, we use `analyzerOptions` directly instead of passing it in as a parameter.

src/main/scala/com/amazon/deequ/analyzers/PatternMatch.scala
```diff
@@ -82,6 +85,19 @@ case class PatternMatch(column: String, pattern: Regex, where: Option[String] =
     conditionalSelection(expression, where).cast(IntegerType)
   }
 
+  private[deequ] def rowLevelResults: Column = {
+    val filteredRowOutcome = getRowLevelFilterTreatment(analyzerOptions)
+    val whereNotCondition = where.map { expression => not(expr(expression)) }
+    val expression = when(regexp_extract(col(column), pattern.toString(), 0) =!= lit(""), 1)
```
rdsharma26, line 91: Can we put `when(regexp_extract(col(column), pattern.toString(), 0) =!= lit(""), 1)` in a common place and reuse it in `criterion` and here?

src/main/scala/com/amazon/deequ/checks/Check.scala
```diff
@@ -1121,7 +1128,28 @@ case class Check(
...
+   * @return
+   */
+  def isContainedIn(
+                     column: String,
+                     allowedValues: Array[String],
+                     assertion: Double => Boolean,
+                     hint: Option[String])
+  : CheckWithLastConstraintFilterable = {
+
+
```
rdsharma26, line 1151: nit: excess whitespace

rdsharma26 (commented):
Great work, @eycho-am !

rdsharma26 (approved):
LGTM

PR #543, opened 2024-03-05
[Min/Max] Apply filtered row behavior at the row level evaluation

rdsharma26:
*Description of changes:*

- This changes from applying the behavior at the analyzer level. It allows us to prevent the usage of MinValue/MaxValue as placeholder values for filtered rows.

### DIFF
--- a/src/main/scala/com/amazon/deequ/constraints/Constraint.scala
+++ b/src/main/scala/com/amazon/deequ/constraints/Constraint.scala
@@ -629,7 +629,9 @@ object Constraint {
     val constraint = AnalysisBasedConstraint[MinState, Double, Double](minimum, assertion,
       hint = hint)
 
-    val sparkAssertion = org.apache.spark.sql.functions.udf(assertion)
+    val updatedAssertion = getUpdatedRowLevelAssertion(assertion, minimum.analyzerOptions)
+    val sparkAssertion = org.apache.spark.sql.functions.udf(updatedAssertion)
+
     new RowLevelAssertedConstraint(
       constraint,
       s"MinimumConstraint($minimum)",
@@ -663,7 +665,9 @@ object Constraint {
     val constraint = AnalysisBasedConstraint[MaxState, Double, Double](maximum, assertion,
       hint = hint)
 
-    val sparkAssertion = org.apache.spark.sql.functions.udf(assertion)
+    val updatedAssertion = getUpdatedRowLevelAssertion(assertion, maximum.analyzerOptions)
+    val sparkAssertion = org.apache.spark.sql.functions.udf(updatedAssertion)
+
     new RowLevelAssertedConstraint(
       constraint,
       s"MaximumConstraint($maximum)",
@@ -916,6 +920,20 @@ object Constraint {
         .getOrElse(0.0)
     }
 
+  private[this] def getUpdatedRowLevelAssertion(assertion: Double => Boolean,
+                                                analyzerOpts: Option[AnalyzerOptions])
+  : java.lang.Double => java.lang.Boolean = {
+    (d: java.lang.Double) => {
+      if (Option(d).isDefined) assertion(d)
+      else analyzerOpts match {
+        case Some(analyzerOptions) => analyzerOptions.filteredRow match {
+          case FilteredRowOutcome.TRUE => true
+          case FilteredRowOutcome.NULL => null
+        }
+        case None => null
+      }
+    }
+  }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -916,6 +920,20 @@ object Constraint {
...
+                                                analyzerOpts: Option[AnalyzerOptions])
+  : java.lang.Double => java.lang.Boolean = {
+    (d: java.lang.Double) => {
+      if (Option(d).isDefined) assertion(d)
+      else analyzerOpts match {
+        case Some(analyzerOptions) => analyzerOptions.filteredRow match {
+          case FilteredRowOutcome.TRUE => true
+          case FilteredRowOutcome.NULL => null
+        }
+        case None => null
```
eycho-am, line 933: We had discussed that the default `analyzerOptions` behavior should be `FilteredRowOutcome.TRUE`, should we modify 933 to be true? (By default filtered rows are `true` instead of `null`.

src/test/scala/com/amazon/deequ/VerificationSuiteTest.scala
```diff
@@ -374,11 +371,11 @@ class VerificationSuiteTest extends WordSpec with Matchers with SparkContextSpec
 
       // filtered rows 1, 2, 3 (where item > 3)
       val minRowLevel = resultData.select(expectedColumn4).collect().map(r => r.getAs[Any](0))
-      assert(Seq(true, true, true, true, true, true).sameElements(minRowLevel))
+      assert(Seq(null, null, null, true, true, true).sameElements(minRowLevel))
```
eycho-am, line 374: these test were written with the intention that without specifying analyzer options, the default behavior would be filtered rows are `true` - related to above comment.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
eycho-am: I was testing out another scenario that may be problematic.
Given the following dataframe:

```
+------+----+----+----+----+-----+-----+
|item  |att1|att2|val1|val2|rule4|rule5|
+------+----+----+----+----+-----+-----+
|1     |a   |f   |1   |1   |true |true |
|22    |b   |d   |2   |NULL|true |true |
|333   |a   |NULL|3   |3   |true |true |
|4444  |a   |f   |4   |4   |true |true |
|55555 |b   |NULL|5   |NULL|true |true |
|666666|a   |f   |6   |6   |true |true |
+------+----+----+----+----+-----+-----+
```

where
```
val analyzerOptions = Option(AnalyzerOptions(filteredRow = FilteredRowOutcome.TRUE))

val min = new Check(CheckLevel.Error, "rule4")
  .hasMin("val2", _ > 0, None, analyzerOptions)
  .where("val1 < 4")
val max = new Check(CheckLevel.Error, "rule5")
  .hasMax("val2", _ < 4, None, analyzerOptions)
  .where("val1 < 4")
```

You'll see that rows 1,2,3 should be skipped -> True
Row 5 should be null as val2 is a null value there.
However, with the above method we convert all nulls to true/null - this doesn't distinguish between null values due to being filtered or null values due to null column values.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -916,6 +920,56 @@ object Constraint {
...
+      }
+
+      def filteredRowOutcome: java.lang.Boolean = {
+        analyzerOptions match {
+          case Some(opts) =>
+            opts.filteredRow match {
+              case FilteredRowOutcome.TRUE => true
+              case FilteredRowOutcome.NULL => null
+            }
+          case None => null
```
eycho-am, line 963: nit: similar comment for the below test-case, should the default behavior for filtered rows be true? Or are we making a decision here that they'll be null by default?

rdsharma26: Based on offline discussion, updated the logic. Thanks @eycho-am for the feedback.

src/test/scala/com/amazon/deequ/VerificationSuiteTest.scala
rdsharma26: Reverted the change.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
rdsharma26: Thanks @eycho-am for the valuable feedback. The latest PR revision contains a new structure for the column that helps maintain the "source" of a row, whether it is in scope and filtered out. That will help in evaluating the correct outcome for each row.

eycho-am (approved):
LGTM

PR #547, opened 2024-03-10
[MinLength/MaxLength] Apply filtered row behavior at the row level evaluation

rdsharma26:
*Description of changes:*

- For certain scenarios, the filtered row behavior for MinLength and MaxLength was not working correctly.
- For example, when using both minLength and maxLength constraints in a single check, and with both using == <value> as an assertion. This was resulting in the row level outcome of the filtered rows to be false. This was because we were replacing values for filtered rows for Min to MaxValue and for Max to MinValue. But a number could not equal both at the same time.

```
val analyzerOptions = AnalyzerOptions(
  nullBehavior = NullBehavior.EmptyString,
  filteredRow = FilteredRowOutcome.TRUE
)

val check = new Check(CheckLevel.Error, "test-check")
  .hasMinLength("Company", _ == 8, analyzerOptions = Some(analyzerOptions)).where("ID > 2")
  .hasMaxLength("Company", _ == 8, analyzerOptions = Some(analyzerOptions)).where("ID > 2")


+---+----------------+-------+-----+-----------+----------+
|ID |Company         |ZipCode|State|City       |test-check|
+---+----------------+-------+-----+-----------+----------+
|1  |Acme            |90210  |CA   |Los Angeles|false     |   <-- Incorrect outcome for filtered row
|2  |Acme            |90211  |CA   |Los Angeles|false     |   <-- Incorrect outcome for filtered row 
|3  |Robocorp        |NULL   |NJ   |NULL       |true      |
|4  |Robocorp        |NULL   |NY   |New York   |true      |
+---+----------------+-------+-----+-----------+----------+


```

- Updated the logic of the row level assertion to MinLength/MaxLength to match what was done for Min/Max.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
@@ -36,6 +36,9 @@ case class MaxLength(column: String, where: Option[String] = None, analyzerOptio
   with FilterableAnalyzer {
 
   override def aggregationFunctions(): Seq[Column] = {
+    // The criterion returns a column where each row contains an array of 2 elements.
+    // The first element of the array is a string that indicates if the row is "in scope" or "filtered" out.
+    // The second element is the value used for calculating the metric. We use "element_at" to extract it.
     max(element_at(criterion, 2).cast(DoubleType)) :: Nil
   }
 
@@ -58,7 +61,7 @@ case class MaxLength(column: String, where: Option[String] = None, analyzerOptio
       case NullBehavior.Fail => when(isNullCheck, Double.MaxValue).otherwise(colLength)
       // Empty String is 0 length string
       case NullBehavior.EmptyString => when(isNullCheck, lit(0.0)).otherwise(colLength)
-      case NullBehavior.Ignore => length(col(column))
+      case NullBehavior.Ignore => colLength
     }
 
     conditionalSelectionWithAugmentedOutcome(updatedColumn, where)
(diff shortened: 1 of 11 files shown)


src/main/scala/com/amazon/deequ/analyzers/MaxLength.scala
```diff
@@ -35,12 +36,12 @@ case class MaxLength(column: String, where: Option[String] = None, analyzerOptio
   with FilterableAnalyzer {
 
   override def aggregationFunctions(): Seq[Column] = {
-    max(criterion) :: Nil
+    max(element_at(criterion, 2).cast(DoubleType)) :: Nil
```
eycho-am, line 39: nit: could we add a comment here explaining this behavior? Think it would be beneficial to clarify that this criterion is returning a column of list/tuples.

src/main/scala/com/amazon/deequ/analyzers/MinLength.scala
```diff
@@ -51,35 +52,16 @@ case class MinLength(column: String, where: Option[String] = None, analyzerOptio
...
-    val colLengths: Column = length(conditionalSelection(column, where)).cast(DoubleType)
-    getNullBehavior match {
-      case NullBehavior.Fail =>
-        conditionSelectionGivenColumn(colLengths, Option(isNullCheck), replaceWith = Double.MinValue)
-      case NullBehavior.EmptyString =>
-        // Empty String is 0 length string
-        conditionSelectionGivenColumn(colLengths, Option(isNullCheck), replaceWith = 0.0).cast(DoubleType)
-      case _ =>
-        colLengths
+    val colLength = length(col(column)).cast(DoubleType)
```
eycho-am, line 56: nice, this is much cleaner

eycho-am (approved):
LGTM - added one small comment for documenting this new behavior

rdsharma26:
Thanks for the review @eycho-am . I've addressed the comments in the latest commit.

eycho-am (approved):

Issue #519, opened 2023-10-30
[BUG] Row based output incorrect when using satisfies check and assertion with upper bound < 1


arsenalgunnershubert777:
**Describe the bug**
When using [satisfies](https://github.com/awslabs/deequ/blob/94821c27eda267560a226b3b82a28d9ca2823033/src/main/scala/com/amazon/deequ/checks/Check.scala#L682-L693) check, the columnar row based output seems unexpected based on the assertion being passed in. This specifically occurs when assertion has bound where upper bound < 1.

**To Reproduce**
1. Create custom check using satisfies, with some sql column condition. 
2. Pass in an assertion function with bounds where the upper bound < 1
3. Run check on input dataframe where some rows pass and some rows fail the column condition.
4. The row based output when calling [rowLevelResultsAsDataFrame](https://github.com/awslabs/deequ/blob/79b0ef7ae372045b381fc262798e0d5972e63fe4/src/main/scala/com/amazon/deequ/VerificationResult.scala#L93) will show all rows as false/fail

Code:
```
Check(CheckLevel.Error, id.value)
        .satisfies(
          sqlColumnCondition,
         "name",
          (d: Double) => d > 0 && d < 1.0
        )
```
Output row based dataframe:
```
+-----+------+------+
|index|values|result|
+-----+------+------+
|    1|  blue| false|
|    2| green| false|
|    3|  blue  false|
|    4|   red| false|
|    5|purple| false|
+-----+------+------+
```
5. However, if the assertion bounds is adjusted where the upper bound < 1.1 (instead of 1), then the row based results look correct

Code:
```
Check(CheckLevel.Error, id.value)
        .satisfies(
          sqlColumnCondition,
         "name",
          (d: Double) => d > 0 && d < 1.1
        )
```
Output row based dataframe (this is correct behavior):
```
+-----+------+------+
|index|values|result|
+-----+------+------+
|    1|  blue|  true|
|    2| green|  true|
|    3|  blue|  true|
|    4|   red| false|
|    5|purple| false|
+-----+------+------+

```
**Expected behavior**
The row based output should show rows that passed and rows that failed based on the columnCondition and shouldn’t be impacted by the assertion. The row based output shouldn’t show every row as false when there are certain rows that passed the columnCondition. The correct example is the one shown directly above.

**Screenshots**
N/A

**Additional context**
[...]

PR #553, opened 2024-04-03
Fix for satisfies row level results bug

rdsharma26:
*Issue #, if available:* https://github.com/awslabs/deequ/issues/519

*Description of changes:*

- The satisfies constraint was incorrectly using the provided assertion to evaluate the row level outcomes. The assertion should only be used to evaluate the final outcome.
- As part of this change, we have updated the row level results to return a true/false. The cast to an integer happens as part of the aggregation result.
- Added a test to verify the row level results using checks made up of different assertions.


mentekid (approved):

eycho-am (approved):
LGTM

PR #555, opened 2024-04-10
Column Count Analyzer and Check

mentekid:
- **Fix flaky KLL test**
- **Move CustomSql state to CustomSql analyzer**
- **Implement new Analyzer to count columns**

*Issue #, if available:*

*Description of changes:*

This adds a dataset-scope analyzer that counts the number of columns in a dataset, and uses the analyzer in a Deequ Check

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
@@ -1,0 +1,5 @@
+package com.amazon.deequ.analyzers
+
+class ColumnCount {
+
+}
--- a/src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
@@ -1,5 +1,66 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ * use this file except in compliance with the License. A copy of the License
+ * is located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ *
+ *
+ */
+
 package com.amazon.deequ.analyzers
 
-class ColumnCount {
+import com.amazon.deequ.metrics.DoubleMetric
+import com.amazon.deequ.metrics.Entity
+import org.apache.spark.sql.DataFrame
 
+case class ColumnCount(where: Option[String] = None) extends Analyzer[NumMatches, DoubleMetric] {
+
+  val name = "ColumnCount"
+  val instance = "*"
+  val entity = Entity.Dataset
+
+
+  /**
+   * Compute the state (sufficient statistics) from the data
+   *
+   * @param data data frame
+   * @return
+   */
+  override def computeStateFrom(data: DataFrame, filterCondition: Option[String]): Option[NumMatches] = {
+    if (filterCondition.isDefined) {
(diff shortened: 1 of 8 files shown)


src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
```diff
@@ -0,0 +1,66 @@
...
+ *
+ */
+
+package com.amazon.deequ.analyzers
+
+import com.amazon.deequ.metrics.DoubleMetric
+import com.amazon.deequ.metrics.Entity
+import org.apache.spark.sql.DataFrame
+
+case class ColumnCount(where: Option[String] = None) extends Analyzer[NumMatches, DoubleMetric] {
```
rdsharma26, line 24: Can we remove `where` ? Or is that a requirement from `Analyzer` ?

src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  val name = "ColumnCount"
+  val instance = "*"
+  val entity = Entity.Dataset
+
+
```
rdsharma26, line 30: nit: Excess whitespace

src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
```diff
@@ -0,0 +1,66 @@
...
+      val numColumns = data.columns.size
+      Some(NumMatches(numColumns))
+    }
+  }
+
+  /**
+   * Compute the metric from the state (sufficient statistics)
+   *
+   * @param state wrapper holding a state of type S (required due to typing issues...)
+   * @return
```
rdsharma26, line 50: nit: Missing scaladoc for `@return`

src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
```diff
@@ -0,0 +1,66 @@
...
+   */
+  override def computeMetricFrom(state: Option[NumMatches]): DoubleMetric = {
+    if (state.isDefined) {
+      Analyzers.metricFromValue(state.get.metricValue(), name, instance, entity)
+    } else {
+      Analyzers.metricFromEmpty(this, name, instance, entity)
+    }
```
rdsharma26, line 57: How about:

```
val metricValue = if(state.isDefined) state.get.metricValue() else this
Analyzers.metricFromEmpty(metricValue, name, instance, entity)
```

src/main/scala/com/amazon/deequ/analyzers/CustomSql.scala
```diff
@@ -26,6 +26,17 @@ import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 
+case class CustomSqlState(stateOrError: Either[Double, String]) extends DoubleValuedState[CustomSqlState] {
```
rdsharma26, line 29: Either is right biased in Scala, which means the right sided value is what you would operate on when doing `map` or `flatMap`. In the case of storing exceptions or error messages, usually they are stored as the left sided value of the Either.

src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
mentekid: removed

src/main/scala/com/amazon/deequ/analyzers/ColumnCount.scala
mentekid: it's `Analyzers.metricFromValue` if defined and `Analyzers.metricFromEmpty` otherwise so this won't work. I'll do a map though.

src/main/scala/com/amazon/deequ/analyzers/CustomSql.scala
mentekid: I'm just moving this from Size.scala to its own class, I can address that in a different PR

shriyavanvari (approved):

shriyavanvari (commented):
LGTM

PR #567, opened 2024-05-06
Tpm/interval strategy

zeotuan:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/scala/com/amazon/deequ/suggestions/rules/FractionalCategoricalRangeRule.scala
+++ b/src/main/scala/com/amazon/deequ/suggestions/rules/FractionalCategoricalRangeRule.scala
@@ -23,16 +23,17 @@ import com.amazon.deequ.metrics.DistributionValue
 import com.amazon.deequ.profiles.ColumnProfile
 import com.amazon.deequ.suggestions.ConstraintSuggestion
 import com.amazon.deequ.suggestions.ConstraintSuggestionWithValue
+import com.amazon.deequ.suggestions.rules.FractionalCategoricalRangeRule.defaultIntervalStrategy
+import com.amazon.deequ.suggestions.rules.interval.{ConfidenceIntervalStrategy, WilsonScoreIntervalStrategy}
 import org.apache.commons.lang3.StringEscapeUtils
-
-import scala.math.BigDecimal.RoundingMode
 
 /** If we see a categorical range for most values in a column, we suggest an IS IN (...)
   * constraint that should hold for most values */
 case class FractionalCategoricalRangeRule(
   targetDataCoverageFraction: Double = 0.9,
   categorySorter: Array[(String, DistributionValue)] => Array[(String, DistributionValue)] =
-    categories => categories.sortBy({ case (_, value) => value.absolute }).reverse
+    categories => categories.sortBy({ case (_, value) => value.absolute }).reverse,
+  intervalStrategy: ConfidenceIntervalStrategy = defaultIntervalStrategy
 ) extends ConstraintRule[ColumnProfile] {
 
   override def shouldBeApplied(profile: ColumnProfile, numRecords: Long): Boolean = {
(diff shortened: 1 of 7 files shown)


zeotuan:
@rdsharma26 Hi can you help review this PR.

src/main/scala/com/amazon/deequ/suggestions/rules/interval/ConfidenceIntervalStrategy.scala
```diff
@@ -0,0 +1,34 @@
...
+    require(0.0 <= confidence && confidence <= 1.0, "confidence must be between 0.0 and 1.0")
+  }
+
+  def calculateZScore(confidence: Double): Double = Gaussian(0, 1)(Rand).inverseCdf(1 - ((1.0 - confidence)/ 2.0))
+}
+
+object ConfidenceIntervalStrategy {
+  val defaultConfidence = 0.95
+
+  case class ConfidenceInterval(lowerBound: Double, upperBound: Double)
```
zeotuan, line 31: Currently also calculate upperBound for these ConfidenceInterval. At the moment we don't actually make use of the upperBound though

rdsharma26:
Thank you for the PR! @zeotuan 
The changes look good to me. Can you fix the failing build ?

One point I would like to discuss is making the Wilson Score Confidence Interval Strategy the new default. Could this potentially break backwards compatibility in terms of behavior? If so, should we stick to the Wald Interval Strategy as the default and update the documentation so that users of Deequ can choose which one they want?

zeotuan:
> Thank you for the PR! @zeotuan The changes look good to me. Can you fix the failing build ?
> 
> One point I would like to discuss is making the Wilson Score Confidence Interval Strategy the new default. Could this potentially break backwards compatibility in terms of behavior? If so, should we stick to the Wald Interval Strategy as the default and update the documentation so that users of Deequ can choose which one they want?

I think making Wilson Score Confidence Interval default right now might introduce some surprising changes to existing data quality pipelines. I will add example usage documentation for this. 
We can potentially introduce plan to change the default in a major version update and add "deprecation" message for now so user migrate themselves.

rdsharma26:
> We can potentially introduce plan to change the default in a major version update and add "deprecation" message for now so user migrate themselves.

This seems like a safe approach. Could we configure the default to be the Wald strategy in the following line?

```
 private val defaultIntervalStrategy: ConfidenceIntervalStrategy = WilsonScoreIntervalStrategy()
 ```

File: src/main/scala/com/amazon/deequ/suggestions/rules/FractionalCategoricalRangeRule.scala [hunk 2]
```diff
@@ -23,16 +23,17 @@ import com.amazon.deequ.metrics.DistributionValue
 import com.amazon.deequ.profiles.ColumnProfile
 import com.amazon.deequ.suggestions.ConstraintSuggestion
 import com.amazon.deequ.suggestions.ConstraintSuggestionWithValue
+import com.amazon.deequ.suggestions.rules.FractionalCategoricalRangeRule.defaultIntervalStrategy
+import com.amazon.deequ.suggestions.rules.interval.{ConfidenceIntervalStrategy, WilsonScoreIntervalStrategy}
```
rdsharma26, line 27: nit: Could we avoid grouped imports and use one import per line?

rdsharma26:
Left a comment and a unit test needs fixing.

src/main/scala/com/amazon/deequ/suggestions/rules/FractionalCategoricalRangeRule.scala
zeotuan: Just to clarify, do we prefer separate import or single import but with each on a single line
```scala
import com.amazon.deequ.suggestions.rules.interval.ConfidenceIntervalStrategy
import com.amazon.deequ.suggestions.rules.interval.WilsonScoreIntervalStrategy
```
or
```scala
import com.amazon.deequ.suggestions.rules.interval{
  ConfidenceIntervalStrategy,
  WilsonScoreIntervalStrategy
}
```

rdsharma26: The former. It helps with automatic resolution of merge conflicts.

zeotuan: Updated

rdsharma26 (approved):
LGTM! Thanks @zeotuan for your continued contribution to Deequ!

PR #577, opened 2024-08-31
Fix performance of building row-level results

### LINKED ISSUE #576: [BUG] Performance for building row-level results scales poorly with number of checks
**Describe the bug**
When generating the row-level results DataFrame using the `VerificationResult::lowLevelResultsAsDataFrame` method, performance grinds to a halt on large numbers of checks (hundreds to thousands):
https://github.com/awslabs/deequ/blob/d495234e4cf2c36e8ca220f156bc21e3f57a3bac/src/main/scala/com/amazon/deequ/VerificationResult.scala#L93-L104

This is due to the `DataFrame::withColumn` method being used iteratively within it. Here's a relevant note from the [`withColumn` docs](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#withColumn(colName:String,col:org.apache.spark.sql.Column):org.apache.spark.sql.DataFrame):
> this method introduces a projection internally. Therefore, calling it multiple times, for instance, via loops in order to add multiple columns can generate big plans which can cause performance issues and even StackOverflowException. To avoid this, use select with the multiple columns at once.

**To Reproduce**
1. Run a verification suite with a large number of checks (300-400 is enough to start feeling the pain on my machine).
2. Generate a row-level result DataFrame using `VerificationResult::rowLevelResultsAsDataFrame`.

The performance issue is with generating the DataFrame, so you don't need to call an action on it.

Here's a test I wrote which shows the issue:
```scala
class LargeChecksResultsTest extends AnyFlatSpec with SparkContextSpec with Inspectors {
  "Deequ" should "handle row-level results for large numbers of checks" in withSparkSession { spark =>
    import spark.implicits._
    val df = Seq("1", "2", "3").toDF("data")

    val lengths = (50 to 400 by 50)
    forAll(lengths) { length =>
      val checks = Seq.tabulate(length)(i =>
        Check(CheckLevel.Error, s"check-$i").isComplete("data"))

      val verificationResult = VerificationSuite()
        .onData(df)
        .addChecks(checks)
        .run()

      println("Gathering row-level results")
      val start = Instant.now()
      val rowLeveLResults = VerificationResult.rowLevelResultsAsDataFrame(spark, verificationResult, df)
      val duration = Duration.between(start, Instant.now())
      println(s"${rowLeveLResults.schema.fields.length} columns in row level results")
      println(s"Duration was ${duration.toMillis}ms")
      println()

      assert(verificationResult.status == CheckStatus.Success)
    }
  }
}
```

**Expected behavior**
Generating the row-level results should be performant for large numbers of checks, at least up to the limit of a verification suite itself.

**Additional context**
Here's the output from running the above test on my machine:
```
Gathering row-level results
51 columns in row level results
Duration was 363ms

Gathering row-level results
101 columns in row level results
Duration was 735ms

Gathering row-level results
151 columns in row level results
Duration was 1641ms

Gathering row-level results
201 columns in row level results
Duration was 3421ms

Gathering row-level results
251 columns in row level results
Duration was 6321ms

Gathering row-level results
301 columns in row level results
Duration was 10691ms

Gathering row-level results
351 columns in row level results
Duration was 16275ms

Gathering row-level results
401 columns in row level results
Duration was 23591ms
```

marcantony:
Fixes #576

Iteratively calling `withColumn` (singular) on a DataFrame causes performance issues when iterating over a large sequence of columns. (See issue for more details.) The code was iterating to add each `(name, column)` pair in a map to the DataFrame, so we can use [`withColumns`](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#withColumns(colsMap:Map%5BString,org.apache.spark.sql.Column%5D):org.apache.spark.sql.DataFrame), which takes a `Map[String, Column]` as its parameter, as a drop-in replacement.

After running the performance test in the bug ticket, the performance is much better:
```
Gathering row-level results
51 columns in row level results
Duration was 67ms

Gathering row-level results
101 columns in row level results
Duration was 53ms

Gathering row-level results
151 columns in row level results
Duration was 49ms

Gathering row-level results
201 columns in row level results
Duration was 45ms

Gathering row-level results
251 columns in row level results
Duration was 48ms

Gathering row-level results
301 columns in row level results
Duration was 48ms

Gathering row-level results
351 columns in row level results
Duration was 41ms

Gathering row-level results
401 columns in row level results
Duration was 41ms
```

### DIFF
--- a/src/main/scala/com/amazon/deequ/VerificationResult.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationResult.scala
@@ -97,7 +97,8 @@ object VerificationResult {
 
     val columnNamesToMetrics: Map[String, Column] = verificationResultToColumn(verificationResult)
 
-    data.withColumns(columnNamesToMetrics)
+    val dataWithID = data.withColumn(UNIQUENESS_ID, monotonically_increasing_id())
+    dataWithID.withColumns(columnNamesToMetrics).drop(UNIQUENESS_ID)
   }
 
   def checkResultsAsJson(verificationResult: VerificationResult,
--- a/src/main/scala/com/amazon/deequ/VerificationResult.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationResult.scala
@@ -97,10 +97,7 @@ object VerificationResult {
 
     val columnNamesToMetrics: Map[String, Column] = verificationResultToColumn(verificationResult)
 
-    val dataWithID = data.withColumn(UNIQUENESS_ID, monotonically_increasing_id())
-    columnNamesToMetrics.foldLeft(dataWithID)(
-      (dataWithID, newColumn: (String, Column)) =>
-        dataWithID.withColumn(newColumn._1, newColumn._2)).drop(UNIQUENESS_ID)
+    data.withColumns(columnNamesToMetrics)
   }
 
   def checkResultsAsJson(verificationResult: VerificationResult,


mentekid (approved):
This looks great - thank you for the in-depth analysis and the fix!

marcantony:
Thanks @mentekid! By the way, any idea how soon a version with this can get published? We're trying to support some users with huge numbers of checks (hundreds to thousands) so I'm hoping to incorporate this in our application soon.

mentekid:
We don't have a set cadence but I think we can kick off a release some time
this week for this and some other changes that have accumulated since our
last one.

marcantony:
Hey @mentekid, just wanted to ask about the release again. Could we maybe get something out early next week?

mentekid:
Hey - what version of spark are you interested in? I think I can kick off the release for 3.5 today, the rest take time as they need separate branches and testing.

marcantony:
Ah, we're actually using the 3.4 branch. No rush to kick things off today anyway because we're not quite ready on our end to fully take advantage of the performance fix yet.

marcantony:
Hey @mentekid, I just realized that `withColumns` was [added in Spark 3.3.0](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#withColumns(colsMap:Map[String,org.apache.spark.sql.Column]):org.apache.spark.sql.DataFrame) so this is going to cause problems for the lower version builds. I'll make a PR today to replace it with a `select` instead.

PR #594, opened 2024-12-16
Fix row level bug when composing outcome

rdsharma26:
*Description of changes:*

- When a check fails due to a precondition failure, the row level results are not evaluated correctly.
- For example, let's say a check has a completeness constraint which passes, and a minimum constraint which fails due to a precondition failure.
- The row level results will be the results for just the completeness constraint. There will be no results generated for the minimum constraint, and therefore the row level results will be incorrect.
- We fix this by adding a default outcome for when the row level result column is not provided by the analyzer.

### DIFF
--- a/src/main/scala/com/amazon/deequ/VerificationResult.scala
+++ b/src/main/scala/com/amazon/deequ/VerificationResult.scala
@@ -147,7 +147,7 @@ object VerificationResult {
       case asserted: RowLevelAssertedConstraint =>
         constraintResult.metric.flatMap(metricToColumn).map(asserted.assertion(_)).orElse(Some(lit(false)))
       case _: RowLevelConstraint =>
-        constraintResult.metric.flatMap(metricToColumn)
+        constraintResult.metric.flatMap(metricToColumn).orElse(Some(lit(false)))
       case _: RowLevelGroupedConstraint =>
         constraintResult.metric.flatMap(metricToColumn)
       case _ => None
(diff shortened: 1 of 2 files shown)


src/main/scala/com/amazon/deequ/VerificationResult.scala
```diff
@@ -144,7 +145,7 @@ object VerificationResult {
     val constraint = constraintResult.constraint
     constraint match {
       case asserted: RowLevelAssertedConstraint =>
-        constraintResult.metric.flatMap(metricToColumn).map(asserted.assertion(_))
+        constraintResult.metric.flatMap(metricToColumn).map(asserted.assertion(_)).orElse(Some(lit(false)))
```
eycho-am, line 148: I probably need to look at the code again, but could you remind me the difference between `RowLevelAssertedConstraint` and `RowLevelConstraint` / why we don't need to add it for the latter? Are `RowLevelConstraint` for checks like Completeness, Uniqueness where row level results are boolean and `RowLevelAssertedConstraint` are for checks like Minimum, Sum where we need to compare the row-level value to a given assertion?

eycho-am:
One clarifying question, but otherwise lgtm

eycho-am (approved):

SamPom100 (approved):

rdsharma26:
Addressed feedback in followup commit.

eycho-am (approved):

Issue #596, opened 2025-01-27
[BUG] AnomalyCheck fails because AnalyzerOptions do not persist in FileSystemMetricsRepository


kchaturvedi:
## **Describe the bug**
When running `AnomalyChecks` via `VerificationSuite` with a `InMemoryMetricsRepository`, and have configured `analyzerOptions` on the `Analyzer`, all checks pass as expected.
However, when simply switching to `FileSystemMetricsRepository`, the same checks fail.

## **To Reproduce**
Given a dataframe such as:
```
val currentData = Seq(
        (1, "red"),
        (2, "blue"),
        (3, "green"),
        (4, "red"),
        (5, "red"),
        (6, "red"),
        (7, "red"),
        (8, "red"),
        (9, "green"),
        (10, "red")
).toDF("index", "values")
```

and given a bare minimum `VerificationSuite` with `analyzerOptions` like:
```
val verificationBuilder = VerificationSuite()
  .onData(currentData)
  .useRepository(metricsRepository)
  .saveOrAppendResult(ResultKey(LocalDateTime.now().toInstant(ZoneOffset.UTC).toEpochMilli))
  .addAnomalyCheck(
    BatchNormalStrategy(
      lowerDeviationFactor = Some(1),
      upperDeviationFactor = Some(1)
    ),
    Compliance(
      s"`values` is red",
      s"`values` IN ('red')",
      analyzerOptions = Some(
        AnalyzerOptions(
          filteredRow = FilteredRowOutcome.NULL
        )
      )
    ),
    None
  )
```

### Using InMemoryMetricsRepository
when using an `InMemoryMetricsRepository`, we can see the results passing:
```
val results = verificationBuilder.run()
assert(results.status == CheckStatus.Success)
results.checkResults.head._2.constraintResults.head
```
```
ConstraintResult(
  AnomalyConstraint(
    Compliance(`values` is red,`values` IN ('red'),None,List(),Some(AnalyzerOptions(Ignore,NULL)))),
    Success,
    None,
    Some(DoubleMetric(Column,Compliance,`values` is red,Success(0.7),Some((values IN (red))))
  )
)
```

### Using FileSystemMetricsRepository
However, when switching to a `FileSystemMetricsRepository`, the same execution fails:
```
results.checkResults.head._2.constraintResults.head
```
```
ConstraintResult(
  AnomalyConstraint(
[...]

PR #597, opened 2025-01-31
Add AnalyzerOptions to Analyzer serialize / deserialize logic

kchaturvedi:
*Issue #, if available:* #596 

*Description of changes:*
This PR addresses a discrepancy when using a `FileSystemMetricsRepository` with an `Analyzer` that has `analyzerOptions` set.

Given an analyzer which has `analyzerOptions` configured:
```
Compliance("rule1", "att1 > 3", columns = List("att1"),
  analyzerOptions = Some(AnalyzerOptions(
    nullBehavior = NullBehavior.Ignore,
    filteredRow = FilteredRowOutcome.NULL
  ))
)
```
Because the `FileSystemMetricsRepository.get()` function [compares the entire analyzer object](https://github.com/awslabs/deequ/blob/310699b434d413f8f07ccf54d6eff9bc4de35e58/src/main/scala/com/amazon/deequ/repository/fs/FileSystemMetricsRepository.scala#L154), it fails to match the current analyzer (with `analyzerOptions`) to any historical metric analyzer.

This issue occurs due to the following:
- When the results are persisted to a `FileSystemMetricsRepository`, the `AnalysisResultSerde.serialize` function [does not add](https://github.com/awslabs/deequ/blob/0476ed582512a10ed2d35d0f8a254fa56e82cd4e/src/main/scala/com/amazon/deequ/repository/AnalysisResultSerde.scala#L248) `analyzerOptions` to the result object.

- Similarly, when `AnalysisResultSerde.deserialize` pulls historic metrics from a file, it [does not parse](https://github.com/awslabs/deequ/blob/0476ed582512a10ed2d35d0f8a254fa56e82cd4e/src/main/scala/com/amazon/deequ/repository/AnalysisResultSerde.scala#L438) `analyzerOptions` even if they were present.

This issue means that no matter how many historical metrics are in the metrics repository file, none would be matched to the current analyzer, which would result in the check failing due to insufficient historical metrics. The current workaround would be to avoid using `analyzerOptions` if using `FileSystemMetricsRepository`, or switch to using `InMemoryMetricsRepository`

The fix I propose in this PR is to add the optional `AnalyzerOptions` object to the serializer and deserializer functions so that they may be persisted and fetched properly. This will enable the `FileSystemMetricsRepository.get()` function to successfully compare two analyzer objects and find an exact match.

Note that there is no issue using an `InMemoryMetricsRepository` as there is no serialization/deserialization happening in that case.

### DIFF
--- a/src/main/scala/com/amazon/deequ/repository/AnalysisResultSerde.scala
+++ b/src/main/scala/com/amazon/deequ/repository/AnalysisResultSerde.scala
@@ -30,7 +30,9 @@ import scala.collection._
 import scala.collection.JavaConverters._
 import java.util.{ArrayList => JArrayList, HashMap => JHashMap, List => JList, Map => JMap}
 import JsonSerializationConstants._
+import com.amazon.deequ.analyzers.FilteredRowOutcome.FilteredRowOutcome
 import com.amazon.deequ.analyzers.Histogram.{AggregateFunction, Count => HistogramCount, Sum => HistogramSum}
+import com.amazon.deequ.analyzers.NullBehavior.NullBehavior
 import org.apache.spark.sql.Column
 import org.apache.spark.sql.functions.expr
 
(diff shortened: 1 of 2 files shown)


eycho-am (approved):
Looks good to me, thanks for raising this.

kchaturvedi:
Thanks so much @eycho-am! When can I expect this to be merged and released?

eycho-am:
Just merged, will comment on this PR again when we release

PR #604, opened 2025-02-19
scaffolding for checking data quality agains DQDL rulesets

happy-coral:
*Issue #, if available:*

*Description of changes:*

We want to add the possibility to check data quality agains DQDL rulesets. For that we need: 

1. parse the DQDL ruleset
2. convert each rule via the registered converters to Deequ checks
3. execute the checks
4. returns results as a Spark DataFrame.

This PR adds first implementation of EvaluateDataQuality where we:

1. introduce a high-level conversion mechanism for mapping DQDL rules to Deequ checks (DQDLRuleTranslator)
2. enable a registration mechanism to plug in new rules (e.g., Completeness, Uniqueness, RowCount).

Note that it is a scaffolding prototype – it lays out the necessary classes and interfaces without full functionality yet


SamPom100 (approved):

src/main/scala/com/amazon/deequ/dqdl/DQDLRuleConverter.scala
```diff
@@ -0,0 +1,38 @@
...
+import software.amazon.glue.dqdl.model.DQRule
+
+trait DQDLRuleConverter {
+  def translate(rule: DQRule): Option[Check]
+}
+
+case class CompletenessRuleConverter() extends DQDLRuleConverter {
+  override def translate(rule: DQRule): Option[Check] = {
+    // todo implement
+    Some(Check(CheckLevel.Error, s"Completeness check: ${rule.getRuleType}"))
```
rdsharma26, line 29: We could implement this by extracting the column from the DQ rule, and passing that to the check.

src/main/scala/com/amazon/deequ/dqdl/DefaultDQDLParser.scala
```diff
@@ -0,0 +1,41 @@
...
+import software.amazon.glue.dqdl.model.DQRuleset
+import software.amazon.glue.dqdl.parser.DQDLParser
+
+import scala.util.{Failure, Success, Try}
+
+trait DQDLParserTrait {
+  def parse(ruleset: String): DQRuleset
+}
+
+object DefaultDQDLParser extends DQDLParserTrait {
```
rdsharma26, line 29: Why do we call it `Default`?

rdsharma26 (approved):

src/main/scala/com/amazon/deequ/dqdl/DefaultDQDLParser.scala
happy-coral: because it provides the default implementation for the DQDLOParserTrait, and the class name  DQDLParser is already taken by the parser itself (from AwsGlueMlDQSpecJavaParserGen-1.0.jar)

Issue #600, opened 2025-02-09
[FEATURE] Avoid Full Scan for Size() Metric Calculation


lawofcycles:
**Is your feature request related to a problem? Please describe.**
Currently, the Size() metric in Deequ is calculated by performing a full scan (i.e., executing a complete .count() on the DataFrame) to determine the total number of rows. This approach is inefficient, particularly on large datasets. Moreover, many other analyzers already compute the row count as part of their aggregations (e.g., for calculating metrics like UniqueValueRatio or Distinctness), which makes an independent Size() calculation redundant and results in unnecessary additional processing.

**Describe the solution you'd like**
I propose that the Size() computation be integrated into the existing aggregation scans performed by other analyzers in the AnalysisRunner. Specifically, when executing scanning analyzers, the row count should be calculated once and reused for the Size() metric rather than triggering an extra full scan. Additionally, for grouping analyzers that derive the total number of rows via frequency aggregations, that information should be repurposed to set the Size() metric. This integration would avoid redundant data scans and improve the overall performance of the analysis process.

**Describe alternatives you've considered**
One alternative is to continue computing the Size() metric as a separate aggregation, but that inherently leads to an extra full scan, which is inefficient. Another option might be to allow users to disable the separate computation via configuration; however, that adds extra complexity and does not fully address the performance issue. Integrating the row count calculation into the existing scan logic seems to be the optimal solution, as it leverages already computed values and minimizes the overall processing overhead.

**Additional context**
There is a [TODO comment in the AnalysisRunner.doAnalysisRun method](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala#L171C1-L175C6) that suggests obtaining the number of rows from other metrics or combining the Size() computation with an extra scan if data is already being processed. Implementing this enhancement would eliminate the redundant full scan for Size() and streamline metric computation, leading to significant performance improvements on large datasets.

eycho-am:
Hi @lawofcycles, thanks for raising this issue and the details you've provided. We'll investigate further and get back to you soon.

lawofcycles:
@eycho-am  Thank you for quick response. I'm interested in working on this enhancement.

eycho-am:
@lawofcycles that'd be awesome. Please feel free to raise a PR and we'll make sure to review it.

lawofcycles:
[...]

PR #605, opened 2025-02-23
Refine row count retrieval to skip redundant Size() scans

lawofcycles:
*Issue #, if available:*
#600 
*Description of changes:*

This PR optimizes the use of Size() metrics in AnalysisRunner to avoid redundant 
Spark scans when obtaining the row count for grouping analyzers. 

Specifically:

- Adds a conditional check to include Size(None) only when:
  1) Grouping analyzers require a global row count (i.e., not a FrequencyBasedAnalyzer).
  2) We haven't already included Size(None) implicitly.
  3) There's at least one analyzer whose state isn't already loaded (so an actual scan is needed).

- Introduces a helper method `actuallyNeedsScanning` to detect whether all 
  required analyzer states are already available from `aggregateWith`. If so, we can skip a new scan entirely.

- Extracts the row count from the Size() metric only if it was actually included in the scanning analyzers, preventing unnecessary computations.

This change addresses the [TODO comment in the AnalysisRunner.doAnalysisRun method](https://github.com/awslabs/deequ/blob/master/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala#L171C1-L175C6) regarding row count retrieval efficiency and reduces unnecessary Spark actions, thereby improving performance.

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/runners/AnalysisRunner.scala
@@ -164,14 +164,50 @@ object AnalysisRunner {
         AnalyzerContext.empty
       }
 
+    // If grouping analyzers exist, we might need the total row count.
+    // If no Size() analyzer is present yet, add Size(None) so that row count
+    // will be computed in the same pass as other scanning analyzers.
+    var scanningAnalyzersToRun = scanningAnalyzers
+
+    // Determines if we have at least one grouping analyzer that requires a global row count.
+    // FrequencyBasedAnalyzer calculates its own row count internally, so it does not need a separate Size().
+    val groupingAnalyzersNeedGlobalCount = groupingAnalyzers.exists { analyzer =>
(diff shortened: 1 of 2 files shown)


lawofcycles:
I am conducting tests to see how these changes lead to improved performance.

lawofcycles:
While this PR does fulfill the TODO about avoiding redundant row-count scans in certain edge cases—particularly if a grouping analyzer already retrieves numRows—in most typical scenarios the performance improvement is quite limited. For instance, I ran tests using TPC-DS at the 3 TB scale factor and saw no noticeable difference in runtime. Deequ already consolidates many scan-based analyzers into a single pass, so the main benefit occurs when users unintentionally include Size() even though a grouping analyzer is sufficient.

That said, the change still tidies up the logic and prevents double-counting in rare cases. I’m happy to leave the decision on merging or closing in your hands—just let me know if you’d like any additional updates or tests!

eycho-am (approved):
Looks good to me, thanks for raising this

PR #646, opened 2025-08-28
Add HistogramBinned analyzer support for custom edges

kyraman:
### Description of changes:

Implement custom edges support for the existing `HistogramBinned` Analyzer. This Analyzer currently only supports equal-width edges (bin count is specified, edges are divided automatically). This adds support to input the edges.

#### Validations:
- There are at least 2 edges. Sorting is not necessary.
- Edges cannot exceed maximum allowed value (1000). Same behavior as existing `Histogram` Analyzer.
- Custom edges and a bin count (which is for equal-width) cannot both be specified.

#### Implementation:
1. Sorts bin edges.
1. Iterates through edges for placement.

#### Notes:
- Null value is counted as a bin (as `Histogram` does). Empty bins also count.
- Bins are left-inclusive and right-exclusive (convention), except for the last bin: [a, b) for regular bins, [a, b] for last bin (includes upper bound).

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/HistogramBinned.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/HistogramBinned.scala
@@ -181,7 +181,8 @@ case class HistogramBinned(
             binDataSeq
           }
 
-          DistributionBinned(finalBins, binCount)
+          val totalBins = finalBins.size // Count null and empty bins
+          DistributionBinned(finalBins, totalBins)
         }
 
         HistogramBinnedMetric(column, value)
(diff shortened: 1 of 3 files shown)


src/main/scala/com/amazon/deequ/analyzers/HistogramBinned.scala
```diff
@@ -99,6 +110,41 @@ case class HistogramBinned(
...
+            numericCol >= edges(i) && numericCol < edges(i + 1)
+          }
+          when(condition, lit(i.toString)).otherwise(lit(Histogram.NullFieldReplacement))
+        } else {
+          // Split bins in half and create decision tree
+          val mid = binIndices.size / 2
+          val (leftBins, rightBins) = binIndices.splitAt(mid)
+          val splitEdge = edges(rightBins.head)
+
+          when(numericCol < splitEdge, buildBinaryCondition(leftBins))
```
happy-coral, line 137: this method creates a binary decision tree structure, calling when().otherwise()  calls build up on the call stack. This makes tail recursion optimization impossible here.  (the recursive call must be the very last operation performed. ) So potentially it will just cause StackOverflow exception. Can we refactor it to be tail recursive or we can just use iterative approach?

happy-coral (commented):
can we simplify this code? it looks like we can compute Edges using the

PR #687, opened 2026-03-26
Require Java 11 and upgrade dqdl to 1.0.7

sudsali:
*Issue #, if available:*

*Description of changes:*

Raises the build to Java 11 and upgrades `dqdl` to 1.0.7.

**Why Java 11 is required.** `dqdl`'s own classes are class file major version 55 from 1.0.2 onward (1.0.0 and 1.0.1 are major 52), so `javac` 8 cannot read them off the compile classpath. The floor comes from the dependency, not from a preference.

- `pom.xml`
  - `maven.compiler.source`/`target` `1.8` → `11`.
  - `dqdl` `1.0.0` → `1.0.7`.
  - `maven-javadoc-plugin` `<source>` `8` → `11`, so the release profile agrees with the compiler. CI never runs `-P release`, so a mismatch here would only surface during a live publish.
- `.github/workflows/maven.yml`, `.github/workflows/publish.yml`: `java-version` `8` → `11`.
- `README.md`: Java requirement scoped by release, since the install snippets still point at `2.0.21-spark-3.5`, which runs on Java 8.
- `src/scripts/generate_kb.py`: the same claim, which `update-kb.yml` publishes to the issue bot's knowledge base on every push to `master`. Also adds `<maven.compiler.` to the pom-excerpt filter — it previously matched only `<java.version>`, which does not exist in this pom, so the generated build-config section never showed a Java level and the hand-written line was the sole source of truth. Now the two corroborate each other.

**Bug fix included.** 1.0.7 is the first `dqdl` whose grammar parses `DuplicateRowCount`, a rule `README.md` has advertised since #727 while `master` pinned 1.0.0. Enabling it exposed a metric-key mismatch in `DuplicateRowCountRule`: the converter labelled the entity `Multicolumn` for any non-empty column list, but the analyzer derives it from arity (`Analyzer.entityFrom`), so the single-column form missed the lookup and returned an empty `EvaluatedMetrics` map. The rule outcome itself is still correct — it is computed by the Deequ `Check`, not by this lookup — so the impact is missing metrics, not a wrong verdict. Fixed to match the existing `UniquenessRule` idiom, which also matches the internal Glue Data Quality translator.

**Tests.** The five `DuplicateRowCount` cases were `ignore {}` placeholders pending exactly this upgrade; their bodies are ported from `release/3.0.3-spark-3.5`. Added a single-column case asserting `EvaluatedMetrics`, which is what catches the bug above.

`mvn clean verify` on JDK 11: **1242 tests pass, 0 failures, 0 ignored** (previously 1236 passing + 5 ignored). `mvn clean verify -P release -DskipTests -Dgpg.skip=true` also passes on JDK 11.

**Notes for review**
- Scala bytecode is unchanged. `maven.compiler.*` governs only `javac`, and `scala-maven-plugin` does not forward it to `scalac`, so 821 of 822 classes stay at major 52 and only `AttributeReferenceCreation` (the one `.java` under `src/main`) becomes 55. Making the jar uniformly 55 is not possible here — `scalac` 2.12.10 supports `-release` only up to 9.
- Java 8 therefore breaks partially: most of Deequ still loads, and `UnsupportedClassVersionError` appears on the `ApproxCountDistinct` path or on any DQDL call.
- `dqdl` 1.0.7 shades the ANTLR runtime and moves `org.antlr:antlr4` to `provided`, which removes six transitive artifacts from Deequ's published graph and resolves the long-standing ANTLR version conflict with Spark 3.5's `antlr4-runtime` 4.9.3.
- The `spark.version` `3.5.0` → `3.5.7` change from the original revision of this PR is dropped; `master` already carries 3.5.7 via #675.

**Known gap, not fixed here.** The no-column form (`Rules=[DuplicateRowCount = 2]`) has the same mismatch and is left as-is. `DeequRulesExecutor.resolveDuplicateRowCountColumns` rebuilds the check with all DataFrame columns but forwards `deequMetricMappings` unchanged, so the mapping stays `Dataset`/`*` while the metric is computed as `Multicolumn.<all,cols>`; the outcome is correct but `EvaluatedMetrics` comes back empty. Resolving it means choosing which key is canonical for that form, which is a separate decision from this upgrade.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -13,10 +13,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up JDK 8
+    - name: Set up JDK 11
       uses: actions/setup-java@v3
       with:
-        java-version: '8'
+        java-version: '11'
         distribution: 'corretto'
         cache: maven
     - name: Build with Maven
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -67,10 +67,10 @@ jobs:
           
           aws secretsmanager get-secret-value --secret-id gpg_private_key --query SecretString --output text > "$KEY_FILE"
 
-      - name: Set up JDK 8
+      - name: Set up JDK 11
         uses: actions/setup-java@6a0805fcefea3d4657a47ac4c165951e33482018 # v4.2.2
         with:
-          java-version: '8'
+          java-version: '11'
           distribution: 'corretto'
           cache: maven
 
@@ -173,10 +173,10 @@ jobs:
         with:
           ref: ${{ matrix.spark == '3.5' && 'master' || format('release/{0}-spark-{1}', inputs.release_version, matrix.spark) }}
 
-      - name: Set up JDK 8
+      - name: Set up JDK 11
         uses: actions/setup-java@6a0805fcefea3d4657a47ac4c165951e33482018 # v4.2.2
         with:
-          java-version: '8'
+          java-version: '11'
           distribution: 'corretto'
 
       - name: Bump version and create PR
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Python users may also be interested in PyDeequ, a Python interface for Deequ. Yo
 
 ## Requirements and Installation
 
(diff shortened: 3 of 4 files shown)


kyraman (approved):

zeotuan:
@sudsali Can we go straight to 3.5.8?

PR #693, opened 2026-04-23
Add Variance analyzer with DQDL rule support and tests

kyraman:
### Description of changes

Add Variance analyzer that computes population variance (σ²) on numeric columns. This is the square of StandardDeviation.

Changes span the full analyzer integration surface:
- **Analyzer**: `Variance.scala` with `VarianceState` (Welford merge)
- **Constraint**: `varianceConstraint()` in `Constraint.scala`
- **Check API**:` hasVariance()` in `Check.scala`
- **DQDL rule**: `VarianceRule` converter + registration in `DQDLRuleTranslator`
- **Serialization**: `AnalysisResultSerde` serialize/deserialize cases
- **State persistence**: `HdfsStateProvider` persist/load for `VarianceState`
- **Profiling**: `ColumnProfiler` computes variance for numeric columns; `NumericColumnProfile` exposes `variance: Option[Double]`

### Testing

- `VarianceTest.scala`:
  - Correct computation on integer data
  - Variance equals stddev² relationship
  - Failure on non-numeric columns
  - Where clause filtering
  - Edge cases: single row (0), all identical values (0), all nulls (None), mixed nulls
  - Type coverage: `DoubleType`, `LongType`, `FloatType`, `DecimalType`
  - Metric metadata (entity, name, instance)
  - State merge correctness across partitions (Welford)
  - `VarianceState` throws on n=0
- Variance added to existing tests: `AnalyzerTests`, `AnalysisTest`, `CheckTest`, `ConstraintsTest`, `NullHandlingTests`, `StateAggregationIntegrationTest`, `StateProviderTest`, `AnalysisResultSerdeTest`, `EvaluateDataQualitySpec`

```
mvn test -Dsuites="com.amazon.deequ.analyzers.VarianceTest"
mvn test
```


src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -829,6 +829,36 @@ object Constraint {
...
+      where: Option[String] = None,
+      hint: Option[String] = None)
+    : Constraint = {
+
+    val variance = Variance(column, where)
+
+    fromAnalyzer(variance, assertion, hint)
+  }
+
+  def fromAnalyzer(
```
SamPom100, line 851: there's another `fromAnalyzer` right above this and they're pretty much the same. could we use a generic method or inline the named constraint creation like how mean / sum / min do?

like
```scala
def varianceConstraint(...): Constraint = {
  val variance = Variance(column, where)
  val constraint = AnalysisBasedConstraint[VarianceState, Double, Double](
    variance, assertion, hint = hint)
  new NamedConstraint(constraint, s"VarianceConstraint($variance)")
}
```

SamPom100 (approved):

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
kyraman: Wait, all of `meanConstraint`, `sumConstraint`, `minConstraint` all have a fromAnalyzer() method. But fair point. Since I am adding some more, let me see what I can do about refactoring in follow-ups.

PR #694, opened 2026-04-23
Add Range analyzer with DQDL rule support and tests

kyraman:
### Description of changes

Add Range analyzer that computes the statistical range (max - min) on numeric columns. Single-pass, scan-shareable. Emits two aggregation functions (`min`, `max`) that Spark evaluates in one scan.

Changes span the full analyzer integration surface:
- **Analyzer**: `Range.scala` with `RangeState` (min/max merge)
- **Constraint**: `rangeConstraint()` in `Constraint.scala`
- **Check API**: `hasRange()` in `Check.scala`
- **DQDL rule**: `RangeRule` converter + registration in `DQDLRuleTranslator`
- **Serialization**: `AnalysisResultSerde` serialize/deserialize cases
- **State persistence**: `HdfsStateProvider` persist/load for `RangeState`
- **Profiling**: `ColumnProfiler` computes range for numeric columns; `NumericColumnProfile` exposes `range: Option[Double]`

### Testing

- `RangeTest.scala`:
  - Correct computation on integer data
  - Range equals max minus min relationship
  - Failure on non-numeric columns
  - Where clause filtering
  - Edge cases: single row (0), all identical values (0), all nulls (None), mixed nulls
  - Type coverage: `DoubleType`, `LongType`, `FloatType`, `DecimalType`
  - Metric metadata (entity, name, instance)
  - State merge correctness across partitions
- Range added to existing tests: `AnalyzerTests`, `AnalysisTest`, `CheckTest`, `ConstraintsTest`, `NullHandlingTests`, `StateProviderTest`, `AnalysisResultSerdeTest`

```
mvn test -Dsuites="com.amazon.deequ.analyzers.RangeTest"
mvn test
```

###### By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


sudsali (changes requested):
Missing `columnsReferenced()` override (added to all `ScanShareableAnalyzer` subclasses in my PR #685). Without it, any batch containing `Range` disables column pruning for V2 DataSource connectors. Same gap exists in `Variance` (#693). Can we fix both together?
The fix can be something like this:

```  
override def columnsReferenced(): Option[Set[String]] =
   if (where.isDefined) None else Some(Set(column))
```

kyraman:
> Missing `columnsReferenced()` override (added to all `ScanShareableAnalyzer` subclasses in my PR #685). Without it, any batch containing `Range` disables column pruning for V2 DataSource connectors. Same gap exists in `Variance` (#693). Can we fix both together? The fix can be something like this:
> 
> ```
> override def columnsReferenced(): Option[Set[String]] =
>    if (where.isDefined) None else Some(Set(column))
> ```

Applied for both, thanks.

sudsali (approved):

PR #707, opened 2026-05-06
fix: reduce PR review false positives, increase context budget

sudsali:
## Changes
  
  - **Fetch full source files** for PR-modified files (3MB cap) — gives the model complete context to avoid false positives about undefined variables/methods
  - **Increase context budget** — `max_context_chars`: 200K→800K, `bedrock_timeout`: 120→240s, `max_tokens`: 4000→8000
  - **Use COMMENT event** instead of REQUEST_CHANGES — stops blocking PR merges for bot suggestions
  - **Log 422 response body** — aids debugging when GitHub review API fails
  
  ## Why
  
  The bot was producing false positives (e.g., claiming `numericCol` is undefined when it's defined outside the diff hunk). Root cause: the model only saw diff context, not full files. Now it gets the complete source of every modified file.
  
  ## Prompts (already deployed via GitHub Secrets)
  
  - Removed dead placeholder sections from issue-classify, issue-respond, followup prompts
  - Added `<context_guide>` to PR review prompt explaining `<full_source_files>`
  - Added anti-false-positive rules (check full source before claiming undefined, don't self-contradict)
  
  ## Testing
  
  - 57/57 unit tests pass
  - Validated against PR #706 false positives — all would be prevented with full source context

### DIFF
--- a/src/scripts/issue_bot/config.py
+++ b/src/scripts/issue_bot/config.py
@@ -33,8 +33,8 @@ class Config:
 
         self.upstream_repo = os.getenv("UPSTREAM_REPO", "awslabs/deequ")
 
-        self.bedrock_timeout = 120
-        self.max_context_chars = 200000
+        self.bedrock_timeout = 240
+        self.max_context_chars = 800000
         self.max_github_search_results = 8
         self.github_api_timeout = 10
         self.allowed_labels = {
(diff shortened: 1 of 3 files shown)


kyraman (approved):

sudsali:
Reply for Comment 1 and 2: Both `get_file_content (line 100)` and `get_pr_files (line 51)` exist in github_client.py. The bot reviewer only saw the diff, not the full file, the exact same false positive pattern we just fixed with <full_source_files>.

Reply for comment 3: `max_context_chars` only governs KB truncation in `knowledge_base.py`, not total system prompt size. The 3MB full-source cap is the context window guard for the PR review path. These are independent budgets for different components, KB rarely needs 800K (deequ KB is 523K), and full sources for a typical PR are 200-400K. Together they fit within the 1M token context window.

Issue #607, opened 2025-04-17
[BUG] isNonNegative check doesn't work correctly with bigint data type


denstern:
**Describe the bug**
Check ```isNonNegative()``` doesn't work correctly cause of conversion problem. This function uses ```satisfies()``` function and in this MR #312 authour decided to use ```DECIMAL(20,10)``` with precision = 20 and scale = 10. Integer digits available to store is ```20 - 10 = 10``` digits, so the maximum value is ```10⁹ - 1 = 999,999,999```. Maximum bigInt value is ```9,223,372,036,854,775,807 (19 digits)``` that's why all digits X > 999,999,999 and X < -999,999,999 are converted to 0.0 with ```COALESCE(CAST($column AS DECIMAL(20,10)), 0.0)``` function

**To Reproduce**
[```isNonNegative()```](https://github.com/awslabs/deequ/blob/e8974556d57fb6fa86c42b7d00930d6d57b1ec5b/src/main/scala/com/amazon/deequ/checks/Check.scala#L1004) uses ```satisfies()``` function, so you may use the expression ```COALESCE(CAST($column AS DECIMAL(20,10)), 0.0)``` and any value insted of $column to see the results

**Expected behavior**
For the bigint data type it's better to convert using ```DECIMAL(38,0)``` expression
For data types using floating-point it's better to create another check with ```decimal``` and ```scale``` parameters in it, e.g. ```isNonNegativeDecimal()```

![Image](https://github.com/user-attachments/assets/b922faf9-97b5-4b96-bd0a-7bcae8138c27)

sudsali:
Closing because the fix (#711) has been merged.

Issue #319, opened 2020-12-03
Bug: Verification suite does not run properly if the column contains spaces with  .isNonNegative.


GuptaTanay:
My file contains spaces in the column names on which I am trying to use deequ. 
I could see that isNonNegative is not running properly when there is a space within the file and it even corrupts other verification functions.

eg. .isNonNegative("helpful_votes") runs fine. But .isNonNegative("helpful votes") gives errors like as below:-

 check              | Data checks                                                                                                                                                          
 check_level        | Error                                                                                                                                                                   
 check_status       | Error                                                                                                                                                                   
 constraint         | CompletenessConstraint(Completeness(Some other column name,None))                                                                                                                   
 constraint_status  | Failure                                                                                                                                                                 
 constraint_message | org.apache.spark.sql.catalyst.parser.ParseException: 
mismatched input 'votes' expecting ')'(line 1, pos 15)

dutrow-qualytics:
relates to #85

sudsali:
Closing because the fix (#711) has been merged.

PR #711, opened 2026-05-13
fix: column escaping for special characters, bigint overflow, and ColumnProfiler performance

sudsali:
## Summary

Fixes three related issues where column names with special characters or large numeric values caused failures in Deequ checks and profiling.

## Issues Fixed

- **#607** — `isNonNegative` check doesn't work correctly with bigint data type
- **#319** — Verification suite fails if the column contains spaces with `.isNonNegative`
- **#396** — `ColumnProfiler` performance degradation with many columns (PR by @rli602)

### 1. `ColumnUtil.escapeColumn` — Comprehensive escaping

**Before:** Only escaped column names containing dots.

**After:** Escapes any column name that isn't a simple identifier (`[a-zA-Z_][a-zA-Z0-9_]*`). Already-escaped names (wrapped in backticks) are returned as-is.

### 2. `Check.scala` — Fixed `isNonNegative`, `isPositive`, and all comparison checks

- All checks now use `escapeColumn(column)` in SQL expressions, fixing failures when column names contain spaces, dots, or other special characters.
- Changed `CAST(column AS DECIMAL(20,10))` to `CAST(column AS DOUBLE)` — `DECIMAL(20,10)` overflows for bigint values larger than 10^10 (e.g., `-99999999999` was silently truncated to a positive value, causing false negatives).
- `DOUBLE` preserves sign correctly for all numeric ranges. Sign detection (`>= 0`, `> 0`) doesn't require decimal precision.

### 3. `ColumnProfiler.castNumericStringColumns` — Performance fix

**Before:** Called `withColumn()` + `drop()` + `withColumnRenamed()` in a loop for each numeric column. Per [Spark documentation](https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.sql.DataFrame.withColumn.html), repeated `withColumn` causes O(n²) plan growth.

**After:** Single `select()` call that casts all columns in one pass. References columns using escaped names to handle dots/spaces correctly.

### 4. `ConstraintSuggestionRunner` — Profile lookup fix

Profile map keys are now raw (unescaped) names for user-facing consistency. The `ConstraintSuggestionRunner` uses `removeEscapeColumn` before looking up profiles, keeping internal processing correct.

## Tests Added

| Test | What it verifies |
|------|-----------------|
| `handle column names with spaces in isNonNegative` | Spaces in column name don't break the check; negative value detected |
| `detect negative bigint values in isNonNegative` | Values like `-99999999999` are correctly identified as negative with DOUBLE cast |
| `handle column names with spaces in isPositive` | Spaces in column name don't break isPositive; zero correctly flagged |
| `handle column names with spaces in comparison checks` | `isLessThan` works with spaced column names |

## Existing Test Coverage

All 1151 tests pass including:
- 69 `CheckTest` tests (was 65)
- 16 `ColumnProfilerTest` tests (covers numeric casting with `select()`)
- 4 `ConstraintSuggestionResultTest` tests (covers dots in column names)
- 1 `SuggestionAndVerificationIntegrationTest` (covers end-to-end with dots)

## Verification

```
mvn clean verify → 1151 succeeded, 0 failed, BUILD SUCCESS
```

### DIFF
--- a/src/main/scala/com/amazon/deequ/utilities/ColumnUtil.scala
+++ b/src/main/scala/com/amazon/deequ/utilities/ColumnUtil.scala
@@ -32,7 +32,7 @@ object ColumnUtil {
     } else if (column.matches("^[a-zA-Z_][a-zA-Z0-9_]*$")) {
       column
     } else {
-      "`" + column + "`"
+      "`" + column.replace("`", "``") + "`"
     }
   }
 
--- a/src/main/scala/com/amazon/deequ/utilities/ColumnUtil.scala
+++ b/src/main/scala/com/amazon/deequ/utilities/ColumnUtil.scala
@@ -27,10 +27,12 @@ object ColumnUtil {
   }
 
   def escapeColumn(column: String): String = {
-    if (column.contains(".")) {
-      "`" + column + "`"
-    } else {
+    if (column.startsWith("`") && column.endsWith("`")) {
       column
+    } else if (column.matches("^[a-zA-Z_][a-zA-Z0-9_]*$")) {
+      column
+    } else {
+      "`" + column + "`"
     }
   }
 
--- a/src/main/scala/com/amazon/deequ/checks/Check.scala
+++ b/src/main/scala/com/amazon/deequ/checks/Check.scala
@@ -37,6 +37,7 @@ import com.amazon.deequ.checks.ColumnCondition.isEachNotNull
 import com.amazon.deequ.constraints.Constraint._
 import com.amazon.deequ.constraints._
 import com.amazon.deequ.metrics.BucketDistribution
+import com.amazon.deequ.utilities.ColumnUtil.escapeColumn
 import com.amazon.deequ.metrics.Distribution
 import com.amazon.deequ.metrics.DistributionBinned
 import com.amazon.deequ.metrics.Metric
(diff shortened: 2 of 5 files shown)


src/main/scala/com/amazon/deequ/utilities/ColumnUtil.scala
```diff
@@ -27,10 +27,12 @@ object ColumnUtil {
...
   def escapeColumn(column: String): String = {
-    if (column.contains(".")) {
-      "`" + column + "`"
-    } else {
+    if (column.startsWith("`") && column.endsWith("`")) {
       column
+    } else if (column.matches("^[a-zA-Z_][a-zA-Z0-9_]*$")) {
+      column
+    } else {
+      "`" + column + "`"
```
joshuazexter: `escapeColumn` doesn't handle column names that contain a backtick. A name like 
```
foo`bar
```
 will be wrapped to 
```
`foo`bar`
```
, which is malformed Spark SQL. The standard convention is to double internal backticks.

joshuazexter (changes requested):

src/main/scala/com/amazon/deequ/utilities/ColumnUtil.scala
joshuazexter:
```suggestion
      "`" + column.replace("`", "``") + "`"
```

joshuazexter (approved):

PR #729, opened 2026-05-26
address review feedback for DuplicateRowCount

awang923:
*Description of changes:*

| Reviewer Comment | Solution |
  |---|---|
  | Missing `columnsReferenced` override — needed for V2 DataSource column pruning | Added `override def columnsReferenced(): Option[Set[String]]` that returns `Some(columns.toSet)` when columns are specified and no WHERE clause |
  | Incomplete serde — `analyzerOptions` dropped on round-trip | Added `analyzerOptions` to both serialization and deserialization in `AnalysisResultSerde` |
  | GroupingAnalyzers change modifies shared behavior unnecessarily | Kept the change (required for AnalysisRunner path which bypasses analyzer overrides) but added comment explaining why other analyzers can't reach it with empty columns (they enforce `atLeastOne` precondition) |
  | `computeStateFrom` creates redundant resolution (runs twice with GroupingAnalyzers change) | Removed `computeStateFrom` override — GroupingAnalyzers is now the single source of empty-column resolution |
  | Three places resolve empty columns — fragile | Consolidated to two: GroupingAnalyzers (for AnalysisRunner path) and `calculate` override (for direct calls + Entity re-wrapping) |
  | WHERE-returns-0 contradicts design doc (says empty metric) | Intended behavior per team decision during design review. Design doc to be updated. |
  | Column ordering guarantee in `getParameters.asScala` | Parser uses `LinkedHashMap` which preserves insertion order — same pattern as `IsUniqueRule` |
  | Missing shared test suite registration | Added tests to AnalyzerTests, CheckTest, NullHandlingTests, StateProviderTest |
  | Missing state merge test | Added — verifies group unique in partition A becomes duplicate after merge with partition B |
  | Missing row-level results test | Added — verifies per-row boolean flags (2 duplicates = true, 2 unique = false) |
  | Missing constraint failure message test | Added — verifies failure message includes actual count |
  | Row-level results not flowing through (implicit from FullColumn trait requirement) | Changed `NamedConstraint` → `RowLevelGroupedConstraint` so per-row duplicate flags appear in `rowLevelResultsAsDataFrame` |
  | No README update | Already present from prior commit |
  | ConstraintRulesTest missing | Not applicable — DuplicateRowCount is dataset-level, not part of constraint suggestion system |

### DIFF
--- a/src/main/scala/com/amazon/deequ/analyzers/DuplicateRowCount.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/DuplicateRowCount.scala
@@ -44,7 +44,7 @@ case class DuplicateRowCount(columns: Seq[String], where: Option[String] = None,
 
   override def fromAggregationResult(result: Row, offset: Int, fullColumn: Option[Column]): DoubleMetric = {
     if (result.isNullAt(offset)) {
-      // WHERE clause matched zero rows — return 0 (no duplicates) instead of empty metric
+      // WHERE clause matched zero rows — return 0 (no duplicates)
       toSuccessMetric(0.0, fullColumn)
     } else {
       val conditionColumn = where.map { expression => expr(expression) }
@@ -78,7 +78,7 @@ case class DuplicateRowCount(columns: Seq[String], where: Option[String] = None,
     }
   }
 
-  /** For empty columns, resolve all DataFrame columns at calculation time */
+  /** For empty columns, resolve all DataFrame columns and re-wrap metric entity */
   override def calculate(
       data: DataFrame,
       aggregateWith: Option[StateLoader] = None,
@@ -104,6 +104,9 @@ case class DuplicateRowCount(columns: Seq[String], where: Option[String] = None,
   }
 
   override def filterCondition: Option[String] = where
+
+  override def columnsReferenced(): Option[Set[String]] =
+    if (columns.isEmpty || where.isDefined) None else Some(columns.toSet)
 }
 
 object DuplicateRowCount {
--- a/src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
+++ b/src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
@@ -67,6 +67,12 @@ object FrequencyBasedAnalyzer {
       where: Option[String] = None)
     : FrequenciesAndNumRows = {
 
+    // Resolve empty groupingColumns to all DataFrame columns.
+    // Only DuplicateRowCount can reach here with empty columns (it overrides preconditions
+    // to skip atLeastOne check). All other analyzers enforce atLeastOne(columnsToGroupOn)
+    // which rejects empty columns before this method is called.
+    require(groupingColumns.nonEmpty || data.columns.nonEmpty,
+      "groupingColumns is empty and DataFrame has no columns")
     val resolvedColumns = if (groupingColumns.isEmpty) data.columns.toSeq else groupingColumns
     val columnsToGroupBy = resolvedColumns.map { name => col(name) }.toArray
     val projectionColumns = columnsToGroupBy :+ col(COUNT_COL)
--- a/src/main/scala/com/amazon/deequ/constraints/Constraint.scala
+++ b/src/main/scala/com/amazon/deequ/constraints/Constraint.scala
@@ -174,7 +174,13 @@ object Constraint {
     val constraint = AnalysisBasedConstraint[FrequenciesAndNumRows, Double, Long](
       duplicateRowCount, assertion, Some(_.toLong), hint)
 
-    new NamedConstraint(constraint, s"DuplicateRowCountConstraint($duplicateRowCount)")
+    if (columns.nonEmpty) {
+      new RowLevelGroupedConstraint(constraint,
+        s"DuplicateRowCountConstraint($duplicateRowCount)",
+        duplicateRowCount.columns)
+    } else {
+      new NamedConstraint(constraint, s"DuplicateRowCountConstraint($duplicateRowCount)")
+    }
   }
 
   /**
--- a/src/main/scala/com/amazon/deequ/dqdl/execution/executors/DeequRulesExecutor.scala
+++ b/src/main/scala/com/amazon/deequ/dqdl/execution/executors/DeequRulesExecutor.scala
@@ -20,8 +20,10 @@ import com.amazon.deequ.{VerificationResult, VerificationSuite}
 import com.amazon.deequ.constraints.{RowLevelAssertedConstraint, RowLevelConstraint, RowLevelGroupedConstraint}
 import com.amazon.deequ.dqdl.execution.DQDLExecutor
 import com.amazon.deequ.dqdl.model.{DeequExecutableRule, DeequMetricMapping, Failed, NoColumn, RuleOutcome, SingularColumn}
+import com.amazon.deequ.dqdl.translation.rules.DuplicateRowCountRule
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.functions.{col, concat, lit}
+import scala.collection.JavaConverters._
 import software.amazon.glue.dqdl.model.DQRule
 
 case class DeequExecutionResult(outcomes: Map[DQRule, RuleOutcome], rowLevelData: DataFrame)
@@ -40,9 +42,13 @@ object DeequRulesExecutor extends DQDLExecutor.RuleExecutor[DeequExecutableRule]
       return DeequExecutionResult(Map.empty, df)
     }
 
+    // Resolve empty DuplicateRowCount columns to all DataFrame columns
+    // so RowLevelGroupedConstraint gets a populated column list for row-level results
+    val resolvedRules = rules.map(rule => resolveDuplicateRowCountColumns(rule, df))
+
     val verificationResult = VerificationSuite()
       .onData(df.toDF())
-      .addChecks(rules.map(_.check))
+      .addChecks(resolvedRules.map(_.check))
       .run()
 
     val rowLevelData = VerificationResult.rowLevelResultsAsDataFrame(
@@ -68,7 +74,7 @@ object DeequRulesExecutor extends DQDLExecutor.RuleExecutor[DeequExecutableRule]
         if (hasRowLevel) Some(check.description) else None
     }.toSet
 
-    val outcomes = rules.map { r =>
+    val outcomes = resolvedRules.map { r =>
       val rowLevelOutcome = if (rowLevelCheckDescriptions.contains(r.check.description)) {
         SingularColumn(r.check.description)
       } else {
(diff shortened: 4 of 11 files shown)


src/main/scala/com/amazon/deequ/analyzers/DuplicateRowCount.scala
```diff
@@ -44,7 +44,7 @@ case class DuplicateRowCount(columns: Seq[String], where: Option[String] = None,
```
kyraman, line 44: > Kept the change (required for AnalysisRunner path which bypasses analyzer overrides) but added comment explaining why other analyzers can't reach it with empty columns (they enforce atLeastOne precondition) 

I don't see **GroupingAnalyzers.scala** in this PR's file list. Either way, my concern isn't about documentation, it's about the change itself. A comment saying "other analyzers can't reach this with empty columns" is relying on a convention ( preconditions) rather than enforcing it. At minimum add a defensive check: `require(groupingColumns.nonEmpty || caller is DuplicateRowCount)` or better, revert it and put the resolution back in `computeStateFrom` where it's scoped to this analyzer only.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
```diff
@@ -174,7 +174,9 @@ object Constraint {
     val constraint = AnalysisBasedConstraint[FrequenciesAndNumRows, Double, Long](
       duplicateRowCount, assertion, Some(_.toLong), hint)
```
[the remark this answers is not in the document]
kyraman: +1, this is a real crash. The DQDL path (`DuplicateRowCount = 0` with no explicit columns) goes through `DuplicateRowCountRule` which passes
`columns = Seq.empty` then hits `.head` here then throws `NoSuchElementException`.

The guard needs to be more than just falling back to `NamedConstraint` though... if we use `NamedConstraint` for the empty-columns case, we lose row-level results entirely for the most common DQDL usage. The calculate override resolves columns at runtime, but the constraint type is decided at construction time before we have a DataFrame.

Options as I see it:
1. Fall back to `NamedConstraint` and accept no row-level results for the all-columns variant (document this limitation)
2. Require at least one column in the DQDL translator (resolve all columns from the schema at translation time, before constraint construction)
3. Defer `RowLevelGroupedConstraint` construction to after column resolution (would need a lazy wrapper or a new constraint subclass)

(1) is the simplest but means the default DQDL syntax (`DuplicateRowCount = 0`) won't produce row-level results... which defeats a core value prop of this feature over CustomSql. I'd go with (2): resolve columns at translation time so `RowLevelGroupedConstraint` always gets a populated column list. Either way, add a test for `DuplicateRowCount(Seq.empty)` through the constraint path.

src/test/scala/com/amazon/deequ/analyzers/DuplicateRowCountTest.scala
```diff
@@ -150,5 +150,46 @@ class DuplicateRowCountTest extends AnyWordSpec with Matchers with SparkContextS
       val result = DuplicateRowCount(Seq("col1", "col2")).calculate(df)
       result.value shouldBe scala.util.Success(0.0)
     }
+
+    "correctly merge state across partitions" in withSparkSession { session =>
```
kyraman, line 154: Good test for the cross-partition case. 

Missing the overlapping-group case: partition A has ("a",1) x2, partition B has ("a",1) x1. After merge the count should be 3 (all three are duplicates). This exercises the outer-join sum logic where both sides have a non-null count for the same key.

src/test/scala/com/amazon/deequ/analyzers/DuplicateRowCountTest.scala
```diff
@@ -150,5 +150,46 @@ class DuplicateRowCountTest extends AnyWordSpec with Matchers with SparkContextS
...
+      val stateA = analyzer.computeStateFrom(dfA).get
+      val stateB = analyzer.computeStateFrom(dfB).get
+      val merged = stateA.sum(stateB)
+
+      val metric = analyzer.computeMetricFrom(Some(merged))
+      // ("a",1) has count 2 after merge -> 2 duplicate rows
+      metric.value shouldBe Success(2.0)
+    }
+
+    "produce correct row-level results" in withSparkSession { session =>
```
kyraman, line 188: This selects by check name (dup-check), which works with `NamedConstraint`. But you switched to `RowLevelGroupedConstraint` in this PR. Does `rowLevelResultsAsDataFrame` still use the check name as the column, or does it switch to the constraint/column name? 

Lets verify this test actually passes with the `RowLevelGroupedConstraint` change, or it might be silently selecting the wrong column.

awang923:
| Comment | Fix |
  |---|---|
  | `RowLevelGroupedConstraint` with `Seq.empty` crashes on `.head` | Guard with `if (columns.nonEmpty)` — uses `RowLevelGroupedConstraint` for explicit columns, falls back to `NamedConstraint` for empty columns |
  | GroupingAnalyzers change relies on convention rather than enforcement | Added `require(groupingColumns.nonEmpty \|\| data.columns.nonEmpty)` guard — fails loudly if any analyzer unexpectedly reaches `computeFrequencies` with empty columns on an empty DataFrame |
  | `computeStateFrom` redundant with GroupingAnalyzers change | Both are needed: `computeStateFrom` handles the `calculate` direct-call path; GroupingAnalyzers handles the `AnalysisRunner` path which bypasses `computeStateFrom`. No double resolution occurs in practice. |
  | Missing overlapping-group state merge test | Added test: partition A has ("a",1) x2, partition B has ("a",1) x1, merged count = 3 |
  | Row-level test — verify column name works with `RowLevelGroupedConstraint` | Test passes: `rowLevelResultsAsDataFrame` uses check description as column name, verified with backtick-quoted select |
  | Missing test for `DuplicateRowCount(Seq.empty)` through constraint path | Added test verifying no `NoSuchElementException` on constraint construction |
  | Missing test for empty columns through VerificationSuite | Added test verifying the AnalysisRunner + GroupingAnalyzers path produces correct metric |

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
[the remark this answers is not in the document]
awang923: fix in rev: Guard with `if (columns.nonEmpty)` — uses `RowLevelGroupedConstraint` for explicit columns, falls back to `NamedConstraint` for empty columns

src/main/scala/com/amazon/deequ/analyzers/DuplicateRowCount.scala
awang923: fix in rev2: Added `require(groupingColumns.nonEmpty || data.columns.nonEmpty)` guard — fails loudly if any analyzer unexpectedly reaches `computeFrequencies` with empty columns on an empty DataFrame

src/test/scala/com/amazon/deequ/analyzers/DuplicateRowCountTest.scala
awang923: Good call, added test in rev 2

src/test/scala/com/amazon/deequ/analyzers/DuplicateRowCountTest.scala
awang923: Test passes: `rowLevelResultsAsDataFrame` uses check description as column name, verified with backtick-quoted select

awang923: also added test verifying no `NoSuchElementException` on constraint construction

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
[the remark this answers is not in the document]
kyraman: This fixes the crash, but it means `DuplicateRowCount = 0` (the default DQDL syntax, no explicit columns) won't produce row-level results. That's the most common usage and row-level attribution is the core value prop of this feature over CustomSql (it's literally the first bullet in the design doc).

We need to resolve columns at translation time in `DuplicateRowCountRule`. The translator runs after parsing and has access to the DataSource schema. Populate the columns list there so `RowLevelGroupedConstraint` always gets a non-empty list and row-level results work for all paths. I don't think we should ship without RLR for the default syntax.

src/main/scala/com/amazon/deequ/analyzers/GroupingAnalyzers.scala
```diff
@@ -67,6 +67,12 @@ object FrequencyBasedAnalyzer {
       where: Option[String] = None)
     : FrequenciesAndNumRows = {
 
+    // Resolve empty groupingColumns to all DataFrame columns.
```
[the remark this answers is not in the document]
awang923: A 0-column DataFrame is not a realistic use case. You can't have a DataFrame with no columns in Spark (it would fail much earlier in the pipeline). The `require` is a defensive safety net, not something users will hit. The error message is clear enough for debugging if it ever fires.

src/main/scala/com/amazon/deequ/constraints/Constraint.scala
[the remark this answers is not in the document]
awang923: addressed here: https://github.com/awslabs/deequ/pull/729/changes#diff-73466521c90b8b416cea3c72dc70d8ca8a787ae9fcf4954dd30bdb42c0cd18d4R47

src/main/scala/com/amazon/deequ/dqdl/execution/executors/DeequRulesExecutor.scala
```diff
@@ -102,4 +108,41 @@ object DeequRulesExecutor extends DQDLExecutor.RuleExecutor[DeequExecutableRule]
       }
     }.toMap
   }
+
+  /**
+   * Resolves empty columns in DuplicateRowCount rules to all DataFrame columns.
+   * This ensures RowLevelGroupedConstraint gets a populated column list for row-level results.
+   */
+  private def resolveDuplicateRowCountColumns(rule: DeequExecutableRule, df: DataFrame): DeequExecutableRule = {
```
kyraman, line 116: This works but also fragile... it reconstructs the entire `Check`. If `DuplicateRowCountRule.convert()` logic changes later, this won't stay in sync.

Simpler approach is to just replace the analyzer's columns inside the existing check:
```
 private def resolveDuplicateRowCountColumns(rule: DeequExecutableRule, df: DataFrame): DeequExecutableRule = {
   val isDuplicateRowCountNoColumns = rule.dqRule.getRuleType == "DuplicateRowCount" &&
     !rule.dqRule.getParameters.asScala.exists(_._1.startsWith("TargetColumn"))

   if (!isDuplicateRowCountNoColumns) return rule

   val allColumns = df.columns.toSeq
   val resolvedCheck = rule.check.replaceConstraint { constraint =>
     Constraint.duplicateRowCountConstraint(allColumns,
       constraint.assertion, constraint.where, constraint.hint)
   }
   DeequExecutableRule(rule.dqRule, resolvedCheck, rule.deequMetricMappings)
 }
```

If `Check` doesn't expose a `replaceConstraint` method, an even simpler option: resolve columns once in `DuplicateRowCountRule.convert()` by accepting an optional schema parameter, then the executor just passes `df.schema` at translation time. That way there's one source of truth for how the check is built.

Not a blocker what you have is correct and tested (ty for the unit tests!). But flagging it as worth cleaning up before it drifts.

kyraman (approved):

